PR #12, opened 2023-06-12
Implemented Arlan

Kisa75:
The invincibility part of his A6 is not implemented yet, will open an issue and come back to it later


internal/character/arlan/eidolon.go
```diff
@@ -0,0 +1,102 @@
...
+		Listeners: modifier.Listeners{
+			OnLimboWaitHeal: func(mod *modifier.ModifierInstance) bool {
+				mod.Engine().Heal(info.Heal{
+					Targets:  []key.TargetID{mod.Owner()},
+					Source:   mod.Owner(),
+					BaseHeal: info.HealMap{model.HealFormula_BY_TARGET_MAX_HP: 0.25},
+				})
+				mod.Engine().RemoveModifier(mod.Owner(), E4)
+				return true
+			},
```
unleashurgeek, line 49: 1. Before the heal, Arlan's revive immediately dispels all debuffs currently on him (call `DispelStatus` w/o specifying count and set order to `LastAdded`)
2. This heal needs to be deferred to the "revive stage" of combat. To do this you need to do `mod.Engine().InsertAbility(...)`. The priority of this insert should be `CharReviveSelf`
3. Based on the datamine, this is not a Heal but a `SetHP` call (so wont be buffed by any healing bonuses, just does `MaxHP * 0.25`)
4. NIT: `mod.RemoveSelf()` is a bit cleaner

internal/character/arlan/talent.go
```diff
@@ -0,0 +1,49 @@
...
+			OnAdd: func(mod *modifier.ModifierInstance) {
+				addedBonusDamage := (1 - mod.Engine().HPRatio(mod.Owner())) * mod.State().(talentState).maxBonusDamage
+				mod.SetProperty(prop.AllDamagePercent, addedBonusDamage)
+			},
+
+			// update bonus damage based on new HP
+			OnHPChange: func(mod *modifier.ModifierInstance, e event.HPChangeEvent) {
+				addedBonusDamage := (1 - mod.Engine().HPRatio(mod.Owner())) * mod.State().(talentState).maxBonusDamage
+				mod.SetProperty(prop.AllDamagePercent, addedBonusDamage)
+			},
```
unleashurgeek, line 35: nit: event also has the `NewHPRatio` for you to reference, but is fine as-is

unleashurgeek (approved):

PR #14, opened 2023-06-12
Implementation of the 4 star lightcone of the Path of Destruction: Nowhere to Run.

SuskeDoesStuff:
Implemented the 4 star lightcone of the Path of Destruction: Nowhere to Run.


internal/lightcone/destruction/nowheretorun/nowheretorun.go
```diff
@@ -0,0 +1,53 @@
...
+	modifier.Register(key.NowheretoRun, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnTriggerDeath: onTriggerDeath,
+		},
+	})
+}
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	amtATKPercent := 0.24 + 0.06*float64(lc.Ascension)
+	amtHeal := 0.12 + 0.06*float64(lc.Ascension)
```
unleashurgeek, line 37: fix these. In current state, would be `0.3` and `0.18` at ascension 1 which does not match the description

PR #15, opened 2023-06-12
implementation for 5* lightcone: BeforeDawn

Vessv:
Added the lightcone beforedawn.


internal/lightcone/erudition/beforedawn/beforedawn.go
```diff
@@ -0,0 +1,104 @@
...
+//AfterAction if its ult or skill add the SomnusCorpusMod
+func onAfterAction(mod *modifier.ModifierInstance, e event.ActionEvent){
+	if e.AttackType == model.AttackType_ULT ||
+		e.AttackType == model.AttackType_SKILL {
+		amt := mod.State().(float64)
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:   SomnusCorpus,
+			Source: mod.Owner(),
+			State: &somnusState{
+				amt: amt * 2.66666666667,
```
unleashurgeek, line 100: rather than doing this kind of math which will probably end up with precision differences from game, how about we create a custom state struct for the before dawn modifier? That state struct can have 3 values (the value for each modifier) and then can calculate each exactly in the `Create` function

Alternatively, make state = `float64(lc.Ascension)` and do the calcs in each listener

unleashurgeek (approved):

PR #18, opened 2023-06-13
add pela implementation

skippi:


skippi:
Reviewer Note: Are there plans for unit testing character implementations?

internal/character/pela/trace.go
```diff
@@ -0,0 +1,84 @@
...
+func init() {
+	modifier.Register(A2, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeHitAll: func(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+				dmgBoost := 0.0
+				if mod.Engine().ModifierCount(e.Hit.Defender.ID(), model.StatusType_STATUS_DEBUFF) >= 1 {
+					dmgBoost = 0.2
+				}
+				mod.SetProperty(prop.AllDamagePercent, dmgBoost)
+			},
```
unleashurgeek, line 34: In all `OnBeforeHit*` listeners, you need to modify the `Hit` data in the event rather than the modifier itself. `HitStartEvent` contains a snapshot of the `Attacker` and `Defender` stats (`info.Stats`) which is what is used for the hit calculation. Changes to the modifier would only apply to the next hit (and persist beyond this hit call):
```go
if mod.Engine().ModifierCount(e.Hit.Defender.ID(), model.StatusType_STATUS_DEBUFF) >= 1 {
  e.Hit.Attacker.AddProperty(prop.AllDamagePercent, 0.2)
}
```

internal/character/pela/trace.go
```diff
@@ -0,0 +1,84 @@
...
+				mod.SetProperty(prop.EffectHitRate, 0.1)
+			},
+		},
+	})
+
+	modifier.Register(A6, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeHit: func(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+				mod.SetProperty(prop.AllDamagePercent, 0.2)
+			},
```
unleashurgeek, line 50: same comment as other `BeforeHit` listener

internal/character/pela/trace.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if c.info.Traces["1106102"] {
+		for _, char := range c.engine.Characters() {
+			c.engine.AddModifier(char, info.Modifier{
+				Name:   A4,
+				Source: c.id,
+			})
+		}
+	}
```
unleashurgeek, line 82: because this is happening during character creation, there is no guarantee that `c.engine.Characters()` is fully populated with all characters yet (IE: if pela is char1 then this list will be empty). In addition to the current loop, need to also subscribe to the `CharacterAdded` events and add each time that event occurs:
```go
c.engine.Events().CharacterAdded.Subscribe(func(e event.CharacterAddedEvent) {
  c.engine.AddModifier(e.Id, info.Modifier{
    Name:   A4,
    Source: c.id,
  })
})

unleashurgeek (approved):

unleashurgeek (approved):

PR #20, opened 2023-06-13
add silverwolf implementation

skippi:


internal/character/silverwolf/skill.go
```diff
@@ -0,0 +1,119 @@
...
+	types := info.WeaknessMap{}
+	// add team damage types
+	for _, char := range engine.Characters() {
+		info, _ := engine.CharacterInfo(char)
+		types[info.Element] = true
+	}
+	// remove enemy damage types
+	for t := model.DamageType_PHYSICAL; t <= model.DamageType_IMAGINARY; t++ {
+		if engine.Stats(target).IsWeakTo(t) {
+			delete(types, t)
```
unleashurgeek, line 62: nit: `types.Remove(t)` is also a thing (though it just calls `delete`)

internal/character/silverwolf/technique.go
```diff
@@ -0,0 +1,25 @@
...
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: 0.8,
+		},
+		StanceDamage: 0,
+		EnergyGain:   0.0,
+	})
+
+	for _, trg := range c.engine.Enemies() {
+		c.engine.ModifyStance(trg, c.id, 60)
+	}
```
unleashurgeek, line 24: New issue with current sim implementation. Cannot trigger a "Quantum Weakness Break" here. I think the missing piece is `ModifyStance` needs to take in a `DamageType` so that can be included in the `BreakEvent`? depends on how break implementation looks

internal/character/silverwolf/ult.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (c *char) Ult(target key.TargetID, state info.ActionState) {
+	c.engine.AddModifier(target, info.Modifier{
+		Name:     UltDefDown,
+		Source:   c.id,
+		Duration: 3,
+		Chance:   ultChance[c.info.AbilityLevel.Ult-1],
+		Stats:    info.PropMap{prop.DEFPercent: -ultDefDown[c.info.AbilityLevel.Ult-1]},
+	})
```
unleashurgeek, line 31: missing `TickImmediately`

internal/character/silverwolf/eidolon.go
```diff
@@ -0,0 +1,106 @@
...
+
+// After using her Ultimate to attack enemies, Silver Wolf regenerates 7
+// Energy for every debuff that the target enemy currently has. This effect
+// can be triggered up to 5 time(s) in each use of her Ultimate.
+func (c *char) e1(target key.TargetID) {
+	if c.info.Eidolon >= 1 {
+		debuffCount := c.engine.ModifierCount(target, model.StatusType_STATUS_DEBUFF)
+		if debuffCount > 5 {
+			debuffCount = 5
+		}
```
unleashurgeek, line 55: nit: move the `debuffCount` check + limit to ult and pass it in as a param to both e1 and e4 (can wrap the count in `eidolon >= 1` check so count only being done on E1+)

internal/character/silverwolf/technique.go
skippi: I've only observed break being tied directly to the character trigger. So in this case, we could derive the damage type from the included `c.id` parameter (source). Maybe we should look at dm to see if this is a safe work around?

internal/character/silverwolf/eidolon.go
skippi: This was done intentionally as I'm not sure if e4 can proc new debuffs before e1 does its own debuff check. If you can guarantee that it won't add more debuffs, I can go through with this change.

internal/character/silverwolf/skill.go
```diff
@@ -0,0 +1,119 @@
...
+				mod.SetProperty(damageType_ResProperty[dmgType], -0.2)
+			},
+		},
+	})
+}
+
+func chooseWeaknessType(engine engine.Engine, target key.TargetID) (model.DamageType, bool) {
+	types := info.WeaknessMap{}
+	// add team damage types
+	for _, char := range engine.Characters() {
```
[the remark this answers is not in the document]
skippi: Verified by KQM that current impl works as intended. See: https://discord.com/channels/845087716541595668/1101278107312336906/1118251625589506118

internal/character/silverwolf/technique.go
skippi: Left a note here for future implementation: #24

PR #25, opened 2023-06-13
Implemented Bronya

Kisa75:
TODO: Check for EnergyGain on E4 followup attack


internal/character/bronya/ult.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, trg := range targets {
+		c.engine.AddModifier(trg, info.Modifier{
+			Name:   Ult,
+			Source: c.id,
+			State: ultState{
+				bonusAtkPerc:     ultAtkPerc[c.info.AbilityLevel.Ult-1],
+				bonusCDmgDefault: ultCDmgDefault[c.info.AbilityLevel.Ult-1],
+				bonusCDmgBronya:  ultCDmgBronya[c.info.AbilityLevel.Ult-1] * c.engine.Stats(c.id).CritDamage(),
+			},
+			TickImmediately: true,
```
unleashurgeek, line 48: * similar to other comment, can get rid of the `OnAdd` + `state` and instead just construct `Stats` directly here
* Move the `c.engine.Stats(c.id).CritDamage()` call to outside the for loop. `Stats` call is "expensive" and want to only do it once. This is also important because game snapshots her crit so she doesn't buff herself and further increase allies

internal/character/bronya/technique.go
```diff
@@ -1,10 +1,42 @@
...
 	"github.com/simimpact/srsim/pkg/key"
 )
 
+const (
+	Technique key.Modifier = "bronya-technique"
+)
+
+func init() {
+	modifier.Register(Technique, modifier.Config{
+		Listeners: modifier.Listeners{
```
unleashurgeek, line 17: Is a `StatusType == BUFF`

internal/character/bronya/technique.go
```diff
@@ -1,10 +1,42 @@
...
+			Source: c.id,
+		})
+	}
+
+	c.engine.Events().CharacterAdded.Subscribe(func(e event.CharacterAddedEvent) {
+		c.engine.AddModifier(e.Id, info.Modifier{
+			Name:   Technique,
+			Source: c.id,
+		})
+	})
```
unleashurgeek, line 41: this is not necessary here. the `CharacterAdded` subscription was only needed in trace due to that setup happening during `NewInstance`. Inside `Technique`, is guaranteed to have `engine.Characters()` fully populated

internal/character/bronya/ult.go
```diff
@@ -12,39 +12,25 @@ const (
...
 			Name:   Ult,
 			Source: c.id,
-			State: ultState{
-				bonusAtkPerc:     ultAtkPerc[c.info.AbilityLevel.Ult-1],
-				bonusCDmgDefault: ultCDmgDefault[c.info.AbilityLevel.Ult-1],
-				bonusCDmgBronya:  ultCDmgBronya[c.info.AbilityLevel.Ult-1] * c.engine.Stats(c.id).CritDamage(),
-			},
+			Stats: info.PropMap{prop.ATKPercent: ultAtkPerc[c.info.UltLevelIndex()],
+				prop.CritDMG: ultCDmgDefault[c.info.UltLevelIndex()] +
+					ultCDmgBronya[c.info.UltLevelIndex()]*bronyaCDmg},
```
unleashurgeek, line 33: nit: move crit dmg calc before for loop w/ `bronyCDmg` and make this 2 lines:
```go
critBuff := ultCDmgDefault[c.info.UltLevelIndex()] + ultCDmgBronya[c.info.UltLevelIndex()]*bronyaCDmg

...

Stats: info.PropMap{
  prop.ATKPercent: ultAtkPerc[c.info.UltLevelIndex()],
  prop.CritDMG:    critBuff, 
}
```

PR #40, opened 2023-06-16
Implement Only Silence Remains

connor-kounthapanya:


internal/lightcone/hunt/onlysilenceremains/onlysilenceremains.go
```diff
@@ -0,0 +1,72 @@
...
+	if len(engine.Enemies()) <= 2 {
+		engine.AddModifier(owner, info.Modifier{
+			Name:   OnlySilenceRemainsCRBuff,
+			Source: owner,
+			Stats:  info.PropMap{prop.CritChance: amt},
+		})
+	} else {
+		engine.RemoveModifier(owner, OnlySilenceRemainsCRBuff)
+	}
+}
```
unleashurgeek, line 72: the `onTriggerDeath` listener will only run iff the target the modifier is attached to kills. In the current state, `updateCRBuff` will only run if the owner of this LC kills. I would change `onTriggerDeath` to just be an event subscription on `TargetDeath` (so each time a target dies, run the check for if we should try to add the modifier

internal/lightcone/hunt/onlysilenceremains/onlysilenceremains.go
```diff
@@ -0,0 +1,72 @@
...
+	modifier.Register(OnlySilenceRemains, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnTriggerDeath: onTriggerDeath,
+		},
+	})
+
+	modifier.Register(OnlySilenceRemainsCRBuff, modifier.Config{
+		Stacking:   modifier.ReplaceBySource,
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 38: 1. for whatever reason, this modifier is not registered as a `BUFF` in game (so dont want to label it as a buff here)
2. I would leave the `Stacking` unspecified (default is `Unique`). The reason for this is repeat applications when using `ReplaceBySource` will remove the old modifier and add the new version. This will create a lot of noise in the logs, when in reality the modifier add calls will always be identical. Keeping it as `Unique` means `AddModifier` will only add the modifier if it is not already present.

unleashurgeek (approved):

PR #46, opened 2023-06-16
Implemented QQ

Connorhhuang:
Still need to check autarky energy gain, also might have forgotten to implement something because it's 2 am.


internal/character/qingque/skill.go
```diff
@@ -0,0 +1,44 @@
...
+
+const (
+	Skill key.Modifier = "qingque-skill"
+)
+
+func init() {
+	modifier.Register(Skill, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+		Stacking:   modifier.ReplaceBySource,
+		MaxCount:   4,
```
unleashurgeek, line 19: Should also specify a `CountAddWhenStack` (defaults to `0` which means this stack count will never increase)

internal/character/qingque/skill.go
```diff
@@ -0,0 +1,44 @@
...
+		Listeners: modifier.Listeners{
+			OnAdd: skillOnAdd,
+		},
```
unleashurgeek, line 22: missing listener for this buff to get removed at the end of the turn. In dm, they call `RemoveSelf` in `OnPhase2`

internal/character/qingque/skill.go
```diff
@@ -0,0 +1,44 @@
...
+	})
+}
+func (c *char) Skill(target key.TargetID, state info.ActionState) {
+	c.engine.AddModifier(target, info.Modifier{
+		Name:   Skill,
+		Source: c.id,
+	})
+	c.engine.InsertAction(c.id)
+}
```
unleashurgeek, line 31: * missing logic to draw 2 cards?
* after 2 draw, if has 4 of same suit, need to give QQ "hidden hand"/talent modifier?
* Missing E4/Autarky?

internal/character/qingque/skill.go
```diff
@@ -0,0 +1,44 @@
...
+	extraDamage := 0.0
+	instance, err := mod.Engine().CharacterInstance(mod.Owner())
+	if err != nil {
+		// bad stuff idk how to deal with this
+	}
+	c := instance.(*char)
+	mod.Owner()
+	if c.info.Traces["1201102"] {
+		extraDamage = 0.1
+	}
```
unleashurgeek, line 42: rather than using `CharacterInstance` here, it would be better to just pass in the `extraDamage` as part of the state. I imagine 2 ways of doing this:
* pass in a bool for if this trace is enabled
* pass in the `extraDamage` float

internal/character/qingque/trace.go
```diff
@@ -0,0 +1,65 @@
...
+		})
+	}
+}
+
+func A2ActionEndListener(mod *modifier.ModifierInstance, e event.ActionEvent) {
+	instance, err := mod.Engine().CharacterInstance(mod.Owner())
+	if err != nil {
+		// bad stuff idk how to deal with this
+	}
+	c := instance.(*char)
```
unleashurgeek, line 50: you are using the char id (`mod.Owner()`) to get the `c.id`. `c.id == mod.Owner()` in this case since you attach this modifier to the char

internal/character/qingque/trace.go
```diff
@@ -0,0 +1,65 @@
...
+// A2:
+// 	Restores 1 Skill Point when using the Skill. This effect can only be triggered 1 time per battle.
+// A4:
+//	Using the Skill increases DMG Boost effect of attacks by an extra 10%.
+// A6:
+//	Qingque's SPD increases by 10% for 1 turn after using the Enhanced Basic ATK.
+
+func init() {
+	modifier.Register(A2, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
```
unleashurgeek, line 26: this is not registered as a buff in game (from what I can tell, they dont even use a modifier)

internal/character/qingque/trace.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+func (c *char) a6() {
+	if c.info.Traces["1201103"] {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   A6,
+			Source: c.id,
+			Stats:  info.PropMap{prop.SPDPercent: 0.1},
+		})
```
unleashurgeek, line 63: missing duration

internal/character/qingque/ult.go
```diff
@@ -0,0 +1,24 @@
...
+		DamageType: model.DamageType_QUANTUM,
+		AttackType: model.AttackType_ULT,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: ult[c.info.UltLevelIndex()],
+		},
+		StanceDamage: 60.0,
+		EnergyGain:   5,
+	})
+	c.tiles = []int{4, 0, 0}
+	c.suits[0] = "Yu"
```
unleashurgeek, line 22: ult should give the "hidden hand"/talent buff

internal/character/qingque/qingque.go
```diff
@@ -0,0 +1,68 @@
...
+			},
+			Skill: character.Skill{
+				SPNeed:     1,
+				TargetType: model.TargetType_SELF,
+				CanUse: func(engine engine.Engine, instance info.CharInstance) bool {
+					c := instance.(*char)
+					if c.tiles[0] == 4 {
+						return false
+					}
+					return true
```
unleashurgeek, line 33: simplify to `return c.tiles[0] != 4`

internal/character/qingque/qingque.go
```diff
@@ -0,0 +1,68 @@
...
+		engine:      engine,
+		id:          id,
+		info:        charInfo,
+		tiles:       []int{0, 0, 0},
+		suits:       make([]string, 3),
+		unusedSuits: []string{"Wan", "Tong", "Tiao"},
+	}
+	engine.Events().TurnStart.Subscribe(c.talentTurnStartListener)
+
+	return c
```
unleashurgeek, line 67: missing:
```
c.initTraces()
c.initEidolons()
```

internal/character/qingque/attack.go
unleashurgeek, line 1: Lets refactor this file to maximize reusability and make it a bit easier to read. Here are my thoughts onto how:
1. create a `func (c *char) basicAttack(target key.TargetID, isInsert bool)` which will perform the attack for both non-insert and insert basic attacks.
2. create a `func (c *char) enhancedAttack(target key.TargetID, isInsert bool)` for the insert and non-insert enhanced variant
3. reorganize `Attack` to something like the following:
```go
func (c *char) Attack(target key.TargetID, state info.ActionState) {
  atk := c.GetAttack()
  
  // will perform basic or enhanced attack depending on card state
  atk(target, false)

  if c.engine.HasModifier(c.id, Autarky) {
    c.engine.InsertAbility(info.Insert{
      Execute:    func() { atk(target, true) },
      Source:     c.id,
      Priority:   info.CharInsertAttackSelf,
      AbortFlags: []model.BehaviorFlag{
        model.BehaviorFlag_STAT_CTRL,
        model.BehaviorFlag_DISABLE_ACTION
      },
    })
    c.engine.RemoveModifier(c.id, Autarky)
  }

  if c.tiles[0] == 4 {
    // TODO: logic to reset cards to 0?
    c.a6()
    if c.info.Eidolon >= 6 {
      c.engine.ModifySP(1)
    }
    c.engine.RemoveModifier(c.id, Talent)
  } else {
    c.engine.ModifySP(1)
  }
  state.EndAttack()
}

type attackFunc func(target key.TargetID, isInsert bool)

func (c *char) getAttack() attackFunc {
  if c.tiles[0] == 4 {
    return c.enhancedAttack
  }
  return c.basicAttack
}

func (c *char) basicAttack(target key.TargetID, isInsert bool) {
  ...
}

func (c *char) enhancedAttack(target key.TargetID, isInsert bool) {
  ...
}
```

internal/character/qingque/attack.go
```diff
@@ -0,0 +1,105 @@
...
+				Source:     c.id,
+				Priority:   info.CharInsertAttackOthers,
+				AbortFlags: []model.BehaviorFlag{model.BehaviorFlag_STAT_CTRL, model.BehaviorFlag_DISABLE_ACTION},
+			})
+			c.engine.RemoveModifier(c.id, Autarky)
+		}
+		if c.info.Eidolon >= 6 {
+			c.engine.ModifySP(1)
+		}
+		c.engine.RemoveModifier(c.id, Talent)
```
unleashurgeek, line 101: missing logic to set card counts to 0?

internal/character/qingque/trace.go
```diff
@@ -0,0 +1,65 @@
...
+func init() {
+	modifier.Register(A2, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+		Listeners: modifier.Listeners{
+			OnAfterAction: A2ActionEndListener,
+		},
+	})
+	modifier.Register(A6, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 33: missing `STAT_SPEED_UP` behavior flag

internal/character/qingque/attack.go
```diff
@@ -0,0 +1,105 @@
...
+						DamageType: model.DamageType_QUANTUM,
+						AttackType: model.AttackType_INSERT,
+						BaseDamage: info.DamageMap{
+							model.DamageFormula_BY_ATK: atk[c.info.AttackLevelIndex()],
+						},
+						StanceDamage: 30.0,
+					})
+				},
+				Source:     c.id,
+				Priority:   info.CharInsertAttackOthers,
```
unleashurgeek, line 38: Use the `CharInsertAttackSelf` priority

internal/character/qingque/skill.go
```diff
@@ -12,33 +12,48 @@ const (
...
+		},
 	})
+	c.drawTile()
+	c.drawTile()
+	if c.engine.Rand().Float64() < 0.24 {
+		c.engine.AddModifier(target, info.Modifier{
+			Name:   Autarky,
+			Source: c.id,
+		})
+	}
```
unleashurgeek, line 50: this add should also only happen if `c.info.Eidolon >= 4`

internal/character/qingque/card.go
```diff
@@ -0,0 +1,144 @@
...
+func (c *char) discardTile() string {
+	if c.tiles[2] != 0 {
+		c.tiles[2] -= 1
+	} else if c.tiles[1] != 0 {
+		c.tiles[1] -= 1
+	} else if c.tiles[0] != 0 {
+		c.tiles[0] -= 1
+	} else {
+		// This should really never happen
+	}
```
unleashurgeek, line 39: change the `if c.tiles[0] != 0` to be an else case so you don't need the "should never happen" else

Can also clean up a bit with a switch:
```go
switch {
case c.tiles[2] != 0:
  c.tiles[2] -= 1
case c.tiles[1] != 0:
  c.tiles[1] -= 1
default:
  c.tiles[0] -= 1
}
```

internal/character/qingque/card.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (c *char) swap(pos1, pos2 int) {
+	c.tiles[pos1], c.tiles[pos2] = c.tiles[pos2], c.tiles[pos1]
+	c.suits[pos1], c.suits[pos2] = c.suits[pos2], c.suits[pos1]
+}
+
+// Functions if distinct suits aren't needed to save time and memory, for documentation read the comments on the functions including distinct suits
+func (c *char) drawTile() string {
+	c.e2()
+	if c.tiles[0] != 4 {
```
unleashurgeek, line 11: nit: reduce nesting by flipping condition

```go
c.e2()
if c.tiles[0] == 4 {
  return "no draw"
}

// draw logic here
```

internal/character/qingque/card.go
```diff
@@ -0,0 +1,144 @@
...
+		drawn := c.engine.Rand().Intn(3)
+		c.tiles[drawn] += 1
+		if c.tiles[0] >= c.tiles[1] && c.tiles[1] >= c.tiles[2] {
+		} else if c.tiles[1] > c.tiles[0] {
+			c.swap(0, 1)
+		} else if c.tiles[2] > c.tiles[0] {
+			c.swap(0, 2)
+		} else {
+			c.swap(1, 2)
+		}
```
unleashurgeek, line 23: nit: use switch/case where having a blank case is fine (add comment explaining that blank on purpose)

internal/character/qingque/skill.go
```diff
@@ -0,0 +1,59 @@
...
+	c.engine.AddModifier(target, info.Modifier{
+		Name:   Skill,
+		Source: c.id,
+		State: skillState{
+			damageBoost: extraDamage + skill[c.info.SkillLevelIndex()],
+		},
+	})
+	c.drawTile()
+	c.drawTile()
```
unleashurgeek, line 44: If she draws 4 tiles of the same face by her skill, shouldn't her skill give her the "hidden hand" state?

internal/character/qingque/trace.go
```diff
@@ -0,0 +1,61 @@
...
+		})
+	}
+}
+
+func A2ActionEndListener(mod *modifier.ModifierInstance, e event.ActionEvent) {
+	if mod.Owner() == e.Owner && e.Targets[mod.Owner()] {
+		mod.Engine().ModifySP(1)
+		mod.RemoveSelf()
+	}
+}
```
unleashurgeek, line 50: For the `A2`, its only condition should be to restore SP when a skill is used. Shouldn't this just be simplified to just:
```go
if e.AttackType == model.AttackType_SKILL {
  mod.Engine().ModifySP(1)
  mod.RemoveSelf()
}
```

unleashurgeek (approved):

PR #50, opened 2023-06-17
Cornucopia (Light Cone) Implemented

RyanOld:
Cornucopia :
https://starrailstation.com/en/lightcone/20001
Effect : 
When the wearer uses their Skill or Ultimate, their Outgoing Healing increases by (9 + 3*SI )%.
Listeners : 
OnBeforeAction (apply buff) and OnAfterAction (delete buff)
Checker : 
e.AttackType -> model.AttackType_SKILL / _ULT


internal/lightcone/abundance/cornucopia/cornucopia.go
```diff
@@ -0,0 +1,60 @@
...
+func buffHealsOnSkillUlt(mod *modifier.ModifierInstance, e event.ActionEvent) {
+	healAmt := mod.State().(float64)
+	switch e.AttackType {
+	case model.AttackType_SKILL, model.AttackType_ULT :
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:     CornucopiaBuff,
+			Source:   mod.Owner(),
+			Duration: -100,
+			Stats:    info.PropMap{prop.HealBoost: healAmt},
+		})
```
unleashurgeek, line 54: This current logic will re-add the current `CornucopiaBuff` modifier that is already on the target. Since the stacking behavior of this modifier is `Unique`, this will be a no-op (will not replace the existing version of `CornucopiaBuff`).

What you want to do instead is create a 2nd modifier to contain your buff. This is like how they do it in the datamine (the `MEquip_20001_Main` modifier adds `MEquip_20001_HealRatio`). This style of behavior is similar to the Mutual Demise LC which is implemented (in `Create`, adds the `Check` modifier, `Check` modifier adds the `Buff` modifier under certain conditions): https://github.com/simimpact/srsim/blob/main/internal/lightcone/destruction/mutualdemise/mutualdemise.go

Also, for `Duration`, you can just leave it unspecified for modifiers that will have no duration.

internal/lightcone/abundance/cornucopia/cornucopia.go
```diff
@@ -0,0 +1,60 @@
...
+		Path:          model.Path_ABUNDANCE,
+		Promotions:    promotions,
+	})
+	
+	modifier.Register(CornucopiaBuff, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeAction: buffHealsOnSkillUlt,
+			OnAfterAction: removeHealBuff,
+		},
+		Stacking: modifier.Unique,
```
unleashurgeek, line 31: nit: do not need to specify `Stacking` when using the default behavior (`Unique`)

internal/lightcone/abundance/cornucopia/cornucopia.go
```diff
@@ -0,0 +1,60 @@
...
+			Name:     CornucopiaBuff,
+			Source:   mod.Owner(),
+			Duration: -100,
+			Stats:    info.PropMap{prop.HealBoost: healAmt},
+		})
+	}
+}
+//is it neccessary to remove a permanent buff after each turn?
+func removeHealBuff(mod *modifier.ModifierInstance, e event.ActionEvent)  {
+	mod.Engine().RemoveModifier(mod.Owner(), mod.Name())
```
unleashurgeek, line 59: expanding on the above comment, you would want to remove the 2nd `buff` modifier here rather than the `Check` modifier. Removing the check modifier would stop the light cone from working completely

RyanOld:
will simplify, apply the check + buff modifier and fix the buff remover
thanks for the guidance!

PR #55, opened 2023-06-17
Implement Warmth Shortens Cold Nights (LC)

RyanOld:
Warmth Shortens Cold Nights :
Increases the wearer's Max HP by 16%. When using Basic ATK or Skill, restores all allies' HP by an amount equal to 2% of their respective Max HP.


internal/lightcone/abundance/warmthshortenscoldnights/warmthshortenscoldnights.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	WarmthShortensColdNightsHPBuff     key.Modifier = "warmth_shortens_cold_nights_hp_buff"
+	WarmthShortensColdNightsHealCheck  key.Modifier = "warmth_shortens_cold_nights_heal_check"
```
unleashurgeek, line 16: nit: save yourself the typing and just name the variables `Buff` and `Check` so it is easier to use

internal/lightcone/abundance/warmthshortenscoldnights/warmthshortenscoldnights.go
```diff
@@ -0,0 +1,67 @@
...
+	lightcone.Register(key.WarmthShortensColdNights, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_ABUNDANCE,
+		Promotions:    promotions,
+	})
+
+	//hp buff goes here
+	modifier.Register(WarmthShortensColdNightsHPBuff, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
```
unleashurgeek, line 29: The "Increases the wearer's Max HP by X%" part of this is not registered as a `BUFF` in the game. The game is pretty arbitrary about what is/is not a buff. Typically static stuff in light cones like this will never be a `BUFF`

internal/lightcone/abundance/warmthshortenscoldnights/warmthshortenscoldnights.go
```diff
@@ -0,0 +1,67 @@
...
+		Source: owner,
+		Stats:  info.PropMap{prop.HPPercent: amt},
+		State:  amt,
+	})
+	
+	engine.AddModifier(owner, info.Modifier{ 
+		Name:   WarmthShortensColdNightsHealCheck,
+		Source: owner,
+		State:  0.02 + 0.005 * float64(lc.Imposition),
+	})
```
unleashurgeek, line 53: this should be simplified to a single modifier & add modifier call. The first call has the right idea:
1. Create a `Stats` line setting the `HPPercent` prop (will always have this HP increase as long as modifier is attached, which is always)
2. Set `State` to be used in the `healTeamOnBasicOrSkill` for the heal

Note: the heal amount should be `0.015 + 0.005 * float64(lc.Imposition)`

unleashurgeek (approved):

PR #60, opened 2023-06-18
Implement Swordplay

connor-kounthapanya:


internal/lightcone/hunt/swordplay/swordplay.go
```diff
@@ -0,0 +1,89 @@
...
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	engine.AddModifier(owner, info.Modifier{
+		Name:   Check,
+		Source: owner,
+		State:  0.06 + 0.02*float64(lc.Imposition),
+	})
+}
+
+func onBeforeHit(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+	if !mod.Engine().HasModifier(e.Defender, Target) {
```
unleashurgeek, line 60: There is an interesting bug here which is more to do with the current set of tools the `engine` exposes. In the current state, if multiple people have swordplay equipped the targeting information would be "shared" between them. This could create multiple valid targets at the same time. Instead, each character with swordplay should have its own unique target.

The datamine solves this by having their `HasModifier` check also include the `Source`, which is not a check we have in engine. We may want to open an issue to add support for more complex `HasModifier` searches for cases like this.

You could write such a check in this implementation. It would look something like:
```go
func hasModifierFromSource(engine engine.Engine, key key.Modifier, target, source key.TargetID) bool {
  for _, mod := range mod.Engine().GetModifiers(target, key) {
    if mod.Source == source {
      return true
    }
  }
  return false
}
```

internal/lightcone/hunt/swordplay/swordplay.go
```diff
@@ -0,0 +1,89 @@
...
+
+		mod.Engine().AddModifier(e.Defender, info.Modifier{
+			Name:   Target,
+			Source: mod.Owner(),
+		})
+	}
+}
+
+func onAfterHit(mod *modifier.ModifierInstance, e event.HitEndEvent) {
+	mod.Engine().AddModifier(mod.Owner(), info.Modifier{
```
unleashurgeek, line 79: nit: may want to check the stack count and only add if don't have modifier or if stacks < max. The current state works fine, though every hit we will "replace" the mod and log it which may generate a lot of noise when reading logs about what is going on in the sim (every hit does an `AddModifier` of swordplay).

unleashurgeek (approved):

PR #61, opened 2023-06-18
Implement Return to Darkness

connor-kounthapanya:


internal/lightcone/hunt/returntodarkness/returntodarkness.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	ReturntoDarkness key.Modifier = "return_to_darkness"
+)
+
+var wasTriggered bool = false
```
unleashurgeek, line 20: `wasTriggered` needs to be part of the modifier state. With it global, it will be shared between multiple characters that use this LC (+ multiple iterations of the sim running in parallel)

When adding as part of the state, you'll need to pass the state as a pointer reference so you can edit the `wasTriggered` bool. Meshing Cogs LC is an example of this: https://github.com/simimpact/srsim/blob/f1472e5630ca21889a927ca6bdd4ade8cb0afb25/internal/lightcone/harmony/meshingcogs/meshingcogs.go#L36-L54

unleashurgeek (approved):

PR #64, opened 2023-06-19
Implemented 3* LC Shattered Home

SuskeDoesStuff:


internal/lightcone/destruction/shatteredhome/shatteredhome.go
```diff
@@ -0,0 +1,54 @@
...
+
+}
+
+func onBeforeHitAll(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+	if e.Hit.Defender.CurrentHPRatio() > 0.5 {
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:   ShatteredHomeBuff,
+			Source: mod.Owner(),
+			Stats:  info.PropMap{prop.AllDamagePercent: mod.State().(float64)},
+		})
```
unleashurgeek, line 50: You don't want a new buff modifier here. When inside a `BeforeHit` listener, the stats of the attacker and defender has been snapshotted. Any modifiers applied here wont take effect til the next hit.

Inside `BeforeHit` listeners, you instead want to modify the snapshots directly. In dm they modify `Attacker_AllDamageTypeAddedRatio`, so you'd want to modify the attacker's snapshot:
```go
e.Hit.Attacker.AddProperty(prop.AllDamagePercent, mod.State().(float64))
```

This is similar to the modifiers that happen on In The Night LC: https://github.com/simimpact/srsim/blob/3f1f0c81efa576d3834994ca8307c08a7e25a5c2/internal/lightcone/hunt/inthenight/inthenight.go#L70-L74

SuskeDoesStuff: edited accordingly @unleashurgeek

unleashurgeek (approved):

Issue #65, opened 2023-06-19
`OnBeforeDealHeal` needs an AttackType prop


RyanOld:
On the implementation of the Light Cone Post-Op Conversation, it is found that the DM uses the listener `OnBeforeDealHeal` and then check if the source of the heal is an Ultimate.
Current implementation of `OnBeforeDealHeal` does not have `AttackType` prop to do a check, thus need to do a non-faithful implementation using `OnBeforeAction` and `OnAfterAction` (need tc confirmation if both are equivalent).
If future similar instance of this issue exist, might need to add the prop into the event listener.

- [ ] Add AttackType prop to OnBeforeDealHeal Listener

PR #66, opened 2023-06-19
Implemented Post-Op Conversation (LC) (temp)

RyanOld:
LC ID : 21000
Issue : current implementation uses `OnBeforeAction` and `OnAfterAction` to substitute DM implementation of `OnBeforeDealHeal` as related to issue #65 .


internal/lightcone/abundance/postopconversation/postopconversation.go
```diff
@@ -0,0 +1,67 @@
...
+		Listeners: modifier.Listeners{
+			//NOTE : DM uses OnBeforeDealHeal instead of OnBeforeAction. Might need change.
+			OnBeforeAction: buffHealsOnUlt, 
+			OnAfterAction: removeHealBuff,
+		},
+	})
+	//The actual buff modifier goes here
+	modifier.Register(PostOpHeal, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 38: We do not want this registered as a buff because the game does not (if we register this as a buff, will cause other logic to behave different from game like buff count checks).

unleashurgeek (approved):

PR #67, opened 2023-06-19
Implemented Hunter of Glacial Forest

SuskeDoesStuff:


internal/relic/cavern/hunterofglacialforest/hunterofglacialforest.go
```diff
@@ -0,0 +1,63 @@
...
+	modifier.Register(mod, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeAction: onBeforeAction,
+		},
+	})
+
+	modifier.Register(mod, modifier.Config{
+		Stacking:   modifier.ReplaceBySource,
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 50: Need to register under the `buff` modifier key (tests currently failing because you are registering 2 configs to the same modifier name)

SuskeDoesStuff: fixed!

PR #68, opened 2023-06-19
Implemented Eyes of the Prey

sky-larke:


internal/lightcone/nihility/eyesoftheprey/eyesoftheprey.go
```diff
@@ -0,0 +1,37 @@
...
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_NIHILITY,
+		Promotions:    promotions,
+	})
+}
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	ehr_amt := 0.2 + 0.05*float64(lc.Imposition)
+	dot_amt := 0.24 + 0.06*float64(lc.Imposition)
```
unleashurgeek, line 29: need to decrease the base values. In the current state will give `0.25` EHR and `0.3` dot dmg% at imposition lvl 1.
```go
ehr_amt := 0.15 + 0.05*float64(lc.Imposition)
dot_amt := 0.18 + 0.06*float64(lc.Imposition)
```

internal/lightcone/nihility/eyesoftheprey/eyesoftheprey.go
```diff
@@ -0,0 +1,37 @@
...
+
+	engine.AddModifier(owner, info.Modifier{
+		Name: EyesofthePrey,
+		Source: owner,
+		Stats: info.PropMap{prop.EffectHitRate: ehr_amt, prop.DOTDamagePercent: dot_amt},
+		State: Amts{ehr: ehr_amt, dot: dot_amt},
+	})
```
unleashurgeek, line 36: * `State` is only used when you need to pass data from creation to listener logic (like if this modifier does something `OnBeforeHit` or something). Since you are just setting the stats on add, `State` does nothing here
* Should use the `mod` name you declared as a const. With current setup, you are using the lightcone name as the modifier name (though functionally no difference)
* Typically you want to register all the modifiers created; however, in this case registration is not strictly necessary since you do not need anything beyond the blank/default config.

unleashurgeek (approved):

PR #73, opened 2023-06-19
Logging: Implemented logging function

iCarus0712:
Added some basic testing infrastructure and logging functions


pkg/simulator/simulator.go
```diff
@@ -89,6 +91,36 @@ func Run(ctx context.Context, list *gcs.ActionList, cfg *model.SimConfig) (*mode
...
+	})
+	if err != nil {
+		// something went wrong in thread
+		return nil, err
+	}
+	if err2 != nil {
+		//context must have been cancelled
+		return nil, err
+	}
+
```
unleashurgeek, line 123: I would just delete all simulator changes here. This package should just be deleted and functionality merged into `cmd/srsim`, just haven't had the time to do it.

The reality is we do not want the log run to happen unless explicitly requested. In the gcsim ui (refactor) we do not bundle the logs with the results because it would bloat size and instead just provide a seed selector + generate button which will run the log generation (IE: https://simimpact.app/sh/tgrbhNNBPJcJ#tab=sample).

Instead, should add a new `Run` function in `simulation` to be able to run a single iteration with logging enabled. That's what we do in gcsim, where directly in the `cli` code we just do "if asked to generate logs, make the call to generate": https://github.com/genshinsim/gcsim/blob/feature-stats-rework/cmd/gcsim/main.go#L145-L166

For devs + UI development, we'll want the cli to have a "serve" mode where it spins up an http server and makes the logs available so you can just load the UI and view the logs rather than having to parse the json directly. I can do a separate PR to set that up (read: copy/paste from gcsim) and then have cli run the log iteration from there

pkg/engine/event/combat.go
```diff
@@ -15,6 +18,20 @@ type AttackStartEvent struct {
...
+	buf.WriteString("\"Attacker\": " + strconv.Itoa(int(e.Attacker)))
+	buf.WriteString(fmt.Sprintf(",\"Targets\": %v", e.Targets))
+	buf.WriteString(",\"AttackType\": \"" + e.AttackType.String())
+	buf.WriteString("\",\"DamageType\": \"" + e.DamageType.String())
+	buf.WriteByte('"')
+	buf.WriteByte('}')
+	fmt.Println("Hello")
+	fmt.Println(buf.String())
+	return buf.Bytes(), nil
+}
```
unleashurgeek, line 33: from offline discussion, ideally do not want to create a custom `MarshalJSON` for every event. Instead:
1. `logger` should have some generic way to wrap events and add the name
2. rely on nested `json.Marshal` calls rather than modifying bytes directly like here

IE can do something like the following:
```go
type log struct {
  event string      `json:event`
  data  interface{} `json:data`
}

func (l *logger) Log(e interface{}) {
  json.Marshal(log{
    event: <some way to get its name>,
    data:  e,
  })
}
```

By design, `json.Marshal` will call `json.Marshal` on each field value of a struct (marshalling is recursive)

pkg/engine/event/handler/event.go
```diff
@@ -8,3 +8,7 @@ type event interface {
 	// Ex: String()
 	// ToProto()
 }
+
+type cancellableEvent interface {
+	Cancelled()
+}
```
unleashurgeek, line 14: need to fix the cancellable unit tests

pkg/engine/logging/default_logger.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type defaultLogger struct {
+	entries []string
+}
+
+type logEventWrapper struct {
+	EventName string
+	Data      interface{}
+}
```
unleashurgeek, line 16: I wonder for this and all the event payloads if we should add json struct tags and enforce consistent naming + style for fields (camel case?). Does not need to be done as part of this PR.

pkg/engine/logging/logger.go
```diff
@@ -0,0 +1,25 @@
+package logging
+
+var Singleton Logger
+
+type Logger interface {
+	Log(e interface{})
+	Flush() string
```
unleashurgeek, line 7: thinking out loud, should `Flush` even be something in this interface? If I am a test or the cli or wasm, I should know what logger instance I have and what other methods it has, if any. For a solution where we are flushing during the `Log` call, we wouldn't need a `Flush() string` method

pkg/simulation/simulation.go
```diff
@@ -53,6 +54,14 @@ type simulation struct {
 	actionTargets map[key.TargetID]bool
 }
 
+func RunWithLog(cfg *model.SimConfig, eval *eval.Eval, seed int64) (*model.IterationResult, error) {
```
unleashurgeek, line 57: imo `RunWithLog` should take in a `Logger` and perform the singleton assignment rather than relying on the caller to do so

pkg/engine/logging/default_logger.go
```diff
@@ -0,0 +1,43 @@
...
+
+func InitDefaultLogger() {
+	Singleton = &defaultLogger{entries: make([]string, 0, 10)}
+}
```
unleashurgeek, line 20: I think it makes more sense to change this to the following and push `Singleton` assignment to the caller (since this way they also have full access to the specific logger type if there are any extra methods on it that are not defined in the `Logger` interface):
```go
func NewDefaultLogger() *DefaultLogger {
  return &defaultLogger{entries: make([]string, 0, 10)}
}

File: pkg/simulation/simulation.go [hunk 8]
```diff
@@ -53,6 +54,14 @@ type simulation struct {
...
+	cfg.Iterations = 1
```
unleashurgeek, line 58: Should not modify cfg/treat it as read-only. simulation always runs a single iteration, so don't have enough context to set it to 1 here

unleashurgeek (approved):

PR #74, opened 2023-06-20
Early Implementation of Quid Pro Quo (LC)

RyanOld:
code's a bit messy. might need closer review.


RyanOld:
current implementation uses `Engine().Stats()` which is an expensive operation.
subject to change as related to issue #75

unleashurgeek (approved):

PR #77, opened 2023-06-20
Implement Sushang

connor-kounthapanya:


internal/character/sushang/trace.go
```diff
@@ -0,0 +1,113 @@
...
+		mod.Engine().RemoveModifier(mod.Owner(), A2Buff)
+	}
+}
+
+func a4OnBeforeHitAll(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+	if mod.Engine().HasModifier(mod.Owner(), A4Buff) {
+		stacks := mod.Engine().GetModifiers(mod.Owner(), A4Buff)[0].Count
+		e.Hit.Attacker.AddProperty(prop.AllDamagePercent, stacks*0.025)
+	}
+}
```
unleashurgeek, line 87: rather than calling `GetModifiers` (which requires looping over all modifiers attached to the target), lets just pass the current stack count into the state in the `A4Mod`'s `AddModifier` call. Since we call `AddModifier` before each sword-stance hit, this should be fine.

The alternative is to keep this logic as-is and move the `A4Mod` `AddModifier/RemoveModifier` calls to before/after all sword-stance hits. IMO I am fine with either approach. The 2nd approach could help to reduce noise in the logs, but the 1st approach (using `State`) makes it clear/easy to see what stack count is being used in this mod instance.

internal/character/sushang/skill.go
```diff
@@ -0,0 +1,101 @@
...
+		for i := 1; i <= 2; i++ {
+			if isBroken || c.engine.Rand().Float64() < 0.33 {
+				chances[i] = true
+			}
+		}
+	}
+
+	if isBroken || c.engine.Rand().Float64() < 0.33 {
+		chances[0] = true
+	}
```
unleashurgeek, line 47: Looking at dm, the non-ult hit should go **last** (does both ult hits, then full hit)

internal/character/sushang/skill.go
```diff
@@ -0,0 +1,101 @@
...
+		EnergyGain:   0.0,
+		HitRatio:     hitRatio,
+	})
+
+	if c.info.Traces["1206102"] {
+		c.engine.RemoveModifier(c.id, A4Mod)
+		c.a4AddStack()
+	}
+
+	c.e2()
```
unleashurgeek, line 100: nit: `e2` should happen before A4 removes (though I don't think it really matters in the grand scheme of things)?

unleashurgeek (approved):

PR #83, opened 2023-06-23
UI: fading animations for primitives

mnpqraven:
This PR adds fading animations to primitive components that spawn extra DOM nodes on the screen like `Dialog` or `Tooltip`. Since we're going to use a lot of dialogs, smooth exit animations should cover lot of UX and not need to use bloated animation libraries


mnpqraven:
Oh forgot to checkout 'package.json'. Will fix when i get home

PR #84, opened 2023-06-24
Implement Good Night and Sleep Well

sky-larke:


internal/lightcone/nihility/goodnightandsleepwell/goodnightandsleepwell.go
```diff
@@ -0,0 +1,56 @@
...
+		Rarity:        4,
+		Path:          model.Path_NIHILITY,
+		Promotions:    promotions,
+	})
+
+	modifier.Register(Check, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeHitAll: onBeforeHitAll,
+		},
+	})
```
unleashurgeek, line 30: Add `CanModifySnapshot: true` to this config like how it is done in Fermata. This is how we mark modifiers as being usable in snapshots so it can apply to dots (our equivalent of the `OnSnapshotCreate` callback).

internal/lightcone/nihility/goodnightandsleepwell/goodnightandsleepwell.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// For every debuff the target enemy has, the DMG dealt by the wearer increases by 12%/15%/18%/21%/24%,
+// stacking up to 3 time(s). This effect also applies to DoT.
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	stacks := float64(model.StatusType_STATUS_DEBUFF)
+	if stacks > 3 {
+		stacks = 3
+	}
+	amt := 0.09 + 0.03*float64(lc.Imposition)*stacks
```
unleashurgeek, line 40: The stack check needs to happen inside the `onBeforeHitAll` listener. In the current state, a stat check here will only happen when the Light Cone is initially equipped (`Create` is called during sim startup when creating the new character, before any combat starts and when the enemies have not yet been instantiated).

This logic will not get a count of modifiers. `model.StatusType_STATUS_DEBUFF` is just a flag/enum value for how you can search for this data but not the data itself. To get the debuff status count of the target inside of `onbeforeHitAll` you would want to do the following:
```go
func onBeforeHitAll(mod *modifier.ModifierInstance, e event.HitStartEvent) {
  // how many debuffs are on the target being hit
  debuffCount := e.Hit.Defender.StatusCount(model.StatusType_STATUS_DEBUFF)
  ...
}
```

unleashurgeek (approved):

Issue #75, opened 2023-06-20
`Engine().CharacterInfo()` need `.MaxEnergy()` method


RyanOld:
On the implementation of the Light Cone Quid Pro Quo, it is found that one of the condition of this light cone's effect activation is if an ally other than the LC's holder have <50% energy. This calculation can only be made if a character's current energy and their max energy are known.
But currently, the only way to get a character's max energy is by using `Engine().Stats()` which is computationally expensive to use.
Proposal to add in `.MaxEnergy()` method to `Engine().CharacterInfo()`'s first output (`Info.Character` struct specifically).

- [ ] Add `.MaxEnergy()` or `.EnergyRatio()` method to `Info.Character` struct used as `Engine().CharacterInfo()`'s output

PR #85, opened 2023-06-24
Added in `.MaxEnergy()` method to Engine + QPQ apply

RyanOld:
This PR is a fix related to issue #75 where Quid Pro Quo needs access to a char's max energy without using `Engine.Stats()`.
it added in a new method in /attribute/attribute.go inside AttributeGetter interface, and add this functionality into engine.go.
Now we can simply call `Engine.MaxEnergy(char)` to get a char's max energy.
also already replaced it in QPQ implementation.


internal/lightcone/abundance/quidproquo/quidproquo.go
```diff
@@ -45,9 +45,7 @@ func randomlyAddEnergy(mod *modifier.ModifierInstance) {
 	
 	for _, char := range allyList{
 		//check if energy is <50% and current char isn't LC's holder.
-		//NOTE : Engine().CharacterInfo() don't have .MaxEnergy(). 
-		//currently using .Stats which is not efficient.
-		if (mod.Engine().Energy(char) < mod.Engine().Stats(char).MaxEnergy() / 2 && 
+		if (mod.Engine().Energy(char) < mod.Engine().MaxEnergy(char) / 2 &&
```
unleashurgeek, line 48: I suspect in the future we may get other checks like this, but I doubt anything would ever specifically care about the `MaxEnergy` of a character (because that is something fixed and should never change). Might be better to just expose an `EnergyRatio` method instead similar to `HPRatio` we currently have, that way this simplifies to just:
```go
if mod.Engine().EnergyRatio(char) < 0.5 && char != mod.Owner() {
  ... 
}
```

unleashurgeek:
You need to also add a method here (which just passes through the call to attribute) https://github.com/simimpact/srsim/blob/main/pkg/simulation/attribute.go

unleashurgeek (approved):

PR #87, opened 2023-06-25
Implemented basic damage calculation

sky-larke:
First would like to see if what's been done so far is coherent before continuing.


unleashurgeek (commented):
Reminder to also delete my placeholder comments for how to implement since they shouldn't be relevant once the code is fully written

All core systems should have unit tests imo. Will definitely take a bit to think of and write some, more than happy to help with this and split into separate PR or something. Recommend looking at the `modifier` package to see what unit tests look like in golang.

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,65 @@
...
+	res := float64(h.Defender.GetProperty(prop.AllDamageRES))
+	res += h.Defender.GetProperty(prop.DamageRES(h.DamageType)) - h.Attacker.GetProperty(prop.DamagePEN(h.DamageType))
+	res = 1 - res
+
+	if res < -1 {
+		res = -1
+	} else if res > .9 {
+		res = .9
+	}
+	return res
```
unleashurgeek, line 55: the `res` bounds need to be applied **before** you do `(1 - res)`. Consider the scenario where there is no `RES` or `PEN`

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,65 @@
...
+	toughness_multiplier := 0.9
+	if h.Defender.Stance() == 0 {
+		toughness_multiplier = 1
+	}
+
+	fatigue := 1 - h.Attacker.GetProperty(prop.Fatigue)
+	AllDamageReduce := h.Defender.GetProperty(prop.AllDamageReduce)
+	if AllDamageReduce < 0.01 {
+		AllDamageReduce = 0.01
+	}
```
unleashurgeek, line 33: `AllDamageReduce` needs to be `(1 - AllDamageReduce)` when used in formula. In current state, 0% `AllDamageReduce` will result in 0 damage done

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,58 @@
...
+	if crit {
+		crit_dmg += h.Attacker.CritDamage()
+	}
+
+	total := base * dmg * def_mult * res * vul * toughness_multiplier * fatigue * AllDamageReduce * crit_dmg
+	return total
+}
+
+func (mgr *Manager) res(h *info.Hit) float64 {
+	// We don't currently have basic/ult dmg pen/res, dot pen/res, etc. If we do, we need to add it in here.
```
unleashurgeek, line 37: get rid of comment, basic/skill/ult etc stuff is handled by implementation and not by separate stat (the `OnBeforeHit` that modify the properties prior to all this logic)

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,58 @@
...
+		AllDamageReduce = 0.01
+	}
+
+	crit_dmg := 1.0
+	if crit {
+		crit_dmg += h.Attacker.CritDamage()
+	}
+
+	total := base * dmg * def_mult * res * vul * toughness_multiplier * fatigue * AllDamageReduce * crit_dmg
+	return total
```
unleashurgeek, line 33: make these variable names consistent. Should all be camel case

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return res
+}
+
+func (mgr *Manager) vul(h *info.Hit) float64 {
+	vul := 1.0 + h.Defender.GetProperty(prop.AllDamageTaken)
+	vul += h.Attacker.GetProperty(prop.DamageTaken(h.DamageType))
+	if vul > 1.35 {
+		vul = 1.35
+	}
```
unleashurgeek: datamine max: 3.5

pkg/engine/combat/hit.go
```diff
@@ -27,22 +29,82 @@ func (mgr *Manager) performHit(hit *info.Hit) {
...
 		AttackType:       hit.AttackType,
 		DamageType:       hit.DamageType,
-		BaseDamage:       0, // TODO
-		BonusDamage:      0, // TODO
-		TotalDamage:      0, // TODO
-		ShieldDamage:     0, // TODO
-		HPDamage:         0, // TODO
+		HPDamage:         hpUpdate,
+		BaseDamage:       base,
+		BonusDamage:      bonus,
```
unleashurgeek, line 53: nit: should `BonusDamage` be the percentage like it is currently, or `(base * bonus) - base` to represent the additional damage added from the bonus (or something else entirely?)

pkg/engine/combat/hit.go
```diff
@@ -27,22 +29,82 @@ func (mgr *Manager) performHit(hit *info.Hit) {
...
+	for k, v := range dmgMap {
+		switch k {
+		case model.DamageFormula_BY_ATK:
+			damage += v * h.Attacker.ATK()
+		case model.DamageFormula_BY_DEF:
+			damage += v * h.Attacker.DEF()
+		case model.DamageFormula_BY_MAX_HP:
+			damage += v * h.Attacker.MaxHP()
+		case model.DamageFormula_BY_BREAK_DAMAGE:
+			damage += v * breakBaseDamage[h.Attacker.Level()] // TODO: evaluate if this is the best way
```
unleashurgeek, line 78: remove TODO

pkg/engine/combat/hit.go
```diff
@@ -27,22 +29,82 @@ func (mgr *Manager) performHit(hit *info.Hit) {
...
+	// Adds bEffect if ByPureDamage, and dmg% if not.
+	if h.BaseDamage[model.DamageFormula_BY_BREAK_DAMAGE] < 0 {
+		dmg += h.Attacker.GetProperty(prop.AllDamagePercent)
+		dmg += h.Attacker.GetProperty(prop.DamagePercent(h.DamageType))
+		if h.AttackType == model.AttackType_DOT {
+			dmg += h.Attacker.GetProperty(prop.DOTDamagePercent)
+		}
+	} else {
+		dmg += h.Attacker.BreakEffect()
+	}
```
unleashurgeek, line 103: * comment doesn't match logic here (no check for pure damage?)
* checking for if `BY_BREAK_DAMAGE` is not used should be
```go
if _, ok := h.BaseDamage[model.DamageFormula_BY_BREAK_DAMAGE]; !ok {
  ...
}
```

pkg/engine/combat/hit.go
```diff
@@ -3,43 +3,68 @@ package combat
...
+		BonusDamage:         bonus,
+		DefenceMultiplier:   defMult,
+		Resistance:          res,
+		Vulnerability:       vul,
+		ToughnessMultiplier: toughnessMultiplier,
+		Fatigue:             fatigue,
+		AllDamageReduce:     allDamageReduce,
+		CritDamage:          critDmg,
+		TotalDamage:         total,
+		ShieldDamage:        total,
```
unleashurgeek, line 64: `ShieldDamage` should be `total - hpUpdate`

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	return damage
+}
+
+func (mgr *Manager) bonusDamage(h *info.Hit) float64 {
+	dmg := 1.0
+	// If hit doesn't use break damage equation, adds dmg%
+	// Otherwise, adds break effect%
+	if h.AsPureDamage {
```
unleashurgeek, line 31: condition should be flipped (only want dmg% when not pure damage)

pkg/engine/combat/damage.go
```diff
@@ -0,0 +1,100 @@
...
+		dmg += h.Attacker.GetProperty(prop.AllDamagePercent)
+		dmg += h.Attacker.GetProperty(prop.DamagePercent(h.DamageType))
+		if h.AttackType == model.AttackType_DOT {
+			dmg += h.Attacker.GetProperty(prop.DOTDamagePercent)
+		}
+	} else {
```
unleashurgeek, line 37: change this from an else to only apply if using the `ByBreakDamage` formula (so if using break formula + not pure damage, both DMG% and break effect apply)

PR #88, opened 2023-06-26
implement today is another peaceful day

imring:


internal/lightcone/erudition/todayisanotherpeacefulday/todayisanotherpeacefulday.go
```diff
@@ -0,0 +1,53 @@
...
+
+	modifier.Register(mod, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeHit: onBeforeHit,
+		},
+	})
+}
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	maxenergy := engine.Stats(owner).MaxEnergy()
```
unleashurgeek, line 36: nit: better to use `CharacterInfo` instead of `Stats` here

internal/lightcone/erudition/todayisanotherpeacefulday/todayisanotherpeacefulday.go
```diff
@@ -0,0 +1,53 @@
...
+
+	engine.AddModifier(owner, info.Modifier{
+		Name:   mod,
+		Source: owner,
+		Stats:  info.PropMap{prop.AllDamagePercent: amount * maxenergy},
+		State:  amount * maxenergy,
+	})
+}
+
+func onBeforeHit(mod *modifier.ModifierInstance, e event.HitStartEvent) {
```
unleashurgeek, line 50: this listener + adding state to the modifier is unneeded. In the current form, the `Stats` line in the `AddModifier` call will increase the `AllDamagePercent` by the given amount on the character. Then in `OnBeforeHit`, we would be adding another boost of `AllDamagePercent`, effectively doubling the DMG% increase we have applied to this character for all qualified hits

unleashurgeek (approved):

PR #92, opened 2023-06-27
Implement Day One of My New Life [LC]

mnpqraven:


internal/lightcone/preservation/dayoneofmynewlife/dayoneofmynewlife.go
```diff
@@ -0,0 +1,58 @@
...
+func init() {
+	lightcone.Register(key.DayOneofMyNewLife, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_PRESERVATION,
+		Promotions:    promotions,
+	})
+
+	modifier.Register(modaura, modifier.Config{
+		Stacking: modifier.Replace,
```
unleashurgeek, line 31: nit: While the description says "Effects of the same type cannot stack" the stacking behavior in DM is using `ReplaceBySource`, which means multiple copies of this LC buff can stack. Having KQM TC confirm one way or the other, so fine with matching either description or datamine until validated.

internal/lightcone/preservation/dayoneofmynewlife/dayoneofmynewlife.go
```diff
@@ -0,0 +1,58 @@
...
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	// wearer DEF%
+	amt := 0.14 + 0.02*float64(lc.Imposition)
+	engine.AddModifier(owner, info.Modifier{
+		Name:   mod,
+		Source: owner,
+		Stats:  info.PropMap{prop.DEFPercent: amt},
+		State:  amt,
+	})
```
unleashurgeek, line 43: * `State` is unnecessary because this modifier has no listeners and thus nothing that is leveraging the `State` (should only be used when you need to pass information from `AddModifier` -> a listener associated with the modifier).
* When the owner of this LC dies, the buff gets removed from all team members. This modifier should have an `OnBeforeDying` listener which calls `RemoveModifierFromSource` on all characters

internal/lightcone/preservation/dayoneofmynewlife/dayoneofmynewlife.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// allies AllType RES
+	amtAura := 0.07 + 0.01*float64(lc.Imposition)
+	mod := info.Modifier{
+		Name:   modaura,
+		Source: owner,
+		Stats:  info.PropMap{prop.AllDamageRES: amtAura},
+	}
+
+	engine.Events().BattleStart.Subscribe(func(event event.BattleStartEvent) {
```
unleashurgeek, line 53: nit: fine to leave this as-is, but this may have to change in the future when adding support for multiple waves. `BattleStart` happens at the beginning of each wave

internal/lightcone/preservation/dayoneofmynewlife/dayoneofmynewlife.go
```diff
@@ -50,9 +49,21 @@ func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
...
 		}
 	})
+
+	// removes aura on wearer's death
+	engine.Events().TargetDeath.Subscribe(func(event event.TargetDeathEvent) {
+		if event.Target == owner {
+			for _, char := range engine.Characters() {
+				engine.RemoveModifier(char, modaura)
+			}
+		}
```
mnpqraven, line 67: not 100% sure on this event

unleashurgeek (approved):

PR #95, opened 2023-06-27
Implement We Are Wildfire [LC]

mnpqraven:


internal/lightcone/preservation/wearewildfire/wearewildfire.go
```diff
@@ -0,0 +1,58 @@
...
+	// team DMG RES
+	amt := 0.06 + 0.02*float64(lc.Imposition)
+	amtHeal := 0.25 + 0.05*float64(lc.Imposition)
+
+	dmgmod := info.Modifier{
+		Name:     mod,
+		Source:   owner,
+		Stats:    info.PropMap{prop.AllDamageReduce: amt},
+		Duration: 5,
+	}
```
unleashurgeek, line 41: This mod needs to be registered with the following settings:
* Stacking: `ReplaceBySource` (can have multiple if multiple allies have this LC)
* TickMoment: `ModifierPhase1End`
* StatusType: `BUFF` (this can be found by searching the modifier name in the `ExcelOutput/StatusConfig.json` file in the datamine)

internal/lightcone/preservation/wearewildfire/wearewildfire.go
```diff
@@ -0,0 +1,58 @@
...
+
+	heal := info.Heal{
+		Targets: engine.Characters(),
+		Source:  owner,
+		BaseHeal: info.HealMap{
+			model.HealFormula_BY_TARGET_LOST_HP: amtHeal,
+		},
+	}
+
+	engine.Events().BattleStart.Subscribe(func(event event.BattleStartEvent) {
```
unleashurgeek, line 51: same nit comment as other LC about changes when supporting multiple waves

internal/lightcone/preservation/wearewildfire/wearewildfire.go
```diff
@@ -0,0 +1,58 @@
...
+		Duration: 5,
+	}
+
+	heal := info.Heal{
+		Targets: engine.Characters(),
+		Source:  owner,
+		BaseHeal: info.HealMap{
+			model.HealFormula_BY_TARGET_LOST_HP: amtHeal,
+		},
+	}
```
unleashurgeek, line 49: Need to move this `info.Heal` construction into the `Subscribe` function (specifically the `engine.Characters()` call. In the current state, will get the list of characters here, before the `BattleStart` event occurs. Depending on which character has this LC, not all characters will be in `engine.Characters()` return (IE: if this is first char in the party order, this list will be empty since they are created first)

internal/lightcone/preservation/wearewildfire/wearewildfire.go
mnpqraven: would it suffice to mark these events with comments until we tackle adding support for waves ?

unleashurgeek: yeah. And/or just cut an issue and reference all these cases + whatever else gets added to come back to later

internal/lightcone/preservation/wearewildfire/wearewildfire.go
mnpqraven: what do `ModifierPhase1End` and `ModifierPhase2End` denote exactly ?

unleashurgeek: This field is for when modifiers duration ticks down. Atm we only support `ModifierPhase1End` & `ModifierPhase2End` (where if left unspecified modifiers will default to the 2nd). `Phase1` & `Phase2` refer to a specific time in the turn loop which is when stuff happens. `Phase1` is typically where stuff like DoTs deal damage whereas `Phase2` is at the end of the turn after there are no more actions remaining

unleashurgeek (approved):

PR #97, opened 2023-06-27
Implement This Is Me [LC]

mnpqraven:


internal/lightcone/preservation/thisisme/thisisme.go
```diff
@@ -0,0 +1,70 @@
...
+
+// increase ult damage
+func onBeforeHit(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+	state := mod.State().(State)
+	if e.Hit.AttackType == model.AttackType_ULT {
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:   modUlt,
+			Source: mod.Owner(),
+			Stats:  info.PropMap{prop.DEFPercent: state.ultBonus},
+		})
```
mnpqraven, line 63: not 100% sure about this

internal/lightcone/preservation/thisisme/thisisme.go
```diff
@@ -0,0 +1,70 @@
...
+		Stats:  info.PropMap{prop.DEFPercent: amt},
+		State:  State{ultDmg},
+	})
+}
+
+// increase ult damage
+func onBeforeHit(mod *modifier.ModifierInstance, e event.HitStartEvent) {
+	state := mod.State().(State)
+	if e.Hit.AttackType == model.AttackType_ULT {
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
```
unleashurgeek, line 59: this is incorrect. Instead of adding a modifier, in an `OnBeforeHit` listener you want to modify the data within the `Hit` object in the event.

For this LC specifically:
* In the `State` struct you want to add a `map[key.TargetID]bool` which you `make` when you create the struct
* In `OnBeforeHit` you check if attack type is ult && the `hit.Defender.ID()` is not in the map
* if true, add target to the map and then modify the damage data by adding to the `hit.DamageValue` field (`state.ultBonus * hit.Attacker.DEF()`)
* In `OnAfterAction` (get rid of `OnAfterHit`), clear the state map

internal/lightcone/preservation/thisisme/thisisme.go
```diff
@@ -0,0 +1,72 @@
...
+// wearer's Ultimate.
+func init() {
+	lightcone.Register(key.ThisIsMe, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_PRESERVATION,
+		Promotions:    promotions,
+	})
+
+	modifier.Register(modUlt, modifier.Config{
```
unleashurgeek, line 36: listeners should be on `mod` instead of `modUlt` (`modUlt` should just be deleted)

internal/lightcone/preservation/thisisme/thisisme.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if e.Hit.AttackType == model.AttackType_ULT && !hasId {
+		state.idMap[e.Hit.Defender.ID()] = true
+		e.Hit.DamageValue += state.ultBonus * e.Hit.Attacker.DEF()
+	}
+}
+
+// remove modifier so next ult deals ult dmg + only 1x bonus from this lc
+func onAfterAction(mod *modifier.ModifierInstance, e event.ActionEvent) {
+	mod.RemoveSelf()
```
unleashurgeek, line 69: don't want to remove the mod, just clear the list to change the behavior

unleashurgeek (approved):

PR #103, opened 2023-06-28
Implemented Gepard

Kisa75:


internal/character/gepard/attack.go
```diff
@@ -0,0 +1,21 @@
...
+		Source:     c.id,
+		Targets:    []key.TargetID{target},
+		DamageType: model.DamageType_ICE,
+		AttackType: model.AttackType_NORMAL,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: atk[c.info.AttackLevelIndex()],
+		},
+		StanceDamage: 30.0,
+		EnergyGain:   20.0,
+	})
```
unleashurgeek, line 20: this is a 2 hit attack, `HitRatio` is 50/50

internal/character/gepard/eidolon.go
```diff
@@ -0,0 +1,79 @@
...
+	modifier.Register(E4Res, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAdd: func(mod *modifier.ModifierInstance) {
+				mod.SetProperty(prop.EffectRES, 0.2)
+			},
+		},
+	})
+}
+
+func (c *char) e2() {
```
unleashurgeek, line 57: missing eidolon check `c.info.Eidolon >= 2` (dont create subscription unless E2 or greater)

internal/character/gepard/eidolon.go
```diff
@@ -0,0 +1,79 @@
...
+	c.engine.Events().ModifierRemoved.Subscribe(func(event event.ModifierRemovedEvent) {
+		if event.Modifier.Name == common.Freeze && c.engine.HasModifier(event.Target, E2Tracker) {
+			c.engine.AddModifier(event.Target, info.Modifier{
+				Name:     E2,
```
unleashurgeek, line 61: E2 modifier needs to be registered with the following:
- `Stacking`: `Replace`
- `BehaviorFlags`: `model.BehaviorFlag_STAT_SPEED_DOWN`
- `StatusType`: `DEBUFF`

internal/character/gepard/eidolon.go
```diff
@@ -0,0 +1,79 @@
...
+			},
+			OnBeforeDying: func(mod *modifier.ModifierInstance) {
+				targets := mod.Engine().Characters()
+
+				for _, trg := range targets {
+					mod.Engine().RemoveModifier(trg, E4Res)
+				}
+			},
+		},
+	})
```
unleashurgeek, line 46: IMO this logic modifier is not necessary. Instead I would:
* Move all of the `OnAdd` logic to the `e4()` init function
* On the `E4` modifier that is applying res (should just be a singular `gepard-e4` modifier that applies to everyone), give this modifier the `OnBeforeDying` listener. In that modifier, check if the owner is Gepard (gepard is dying). If so, remove the modifier from all allies

internal/character/gepard/eidolon.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	})
+}
+
+func (c *char) e4() {
+	if c.info.Eidolon >= 4 {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   E4,
+			Source: c.id,
+		})
```
unleashurgeek, line 77: This modifier needs to be registered with `StatusType: BUFF`

internal/character/gepard/skill.go
```diff
@@ -0,0 +1,49 @@
...
+		Source:     c.id,
+		Targets:    []key.TargetID{target},
+		DamageType: model.DamageType_ICE,
+		AttackType: model.AttackType_NORMAL,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: skillDMG[c.info.SkillLevelIndex()],
+		},
+		StanceDamage: 60.0,
+		EnergyGain:   30.0,
+	})
```
unleashurgeek, line 25: * 3 hit attack. split: `15/35/50`
* `AttackType: SKILL`
* `EndAttack` immediately after the 3rd hit (freeze happens after attack end)

internal/character/gepard/talent.go
```diff
@@ -0,0 +1,74 @@
...
+type talentState struct {
+	revivePerc float64
+	a4Active   bool
+	e6Active   bool
+}
+
+func init() {
+	modifier.Register(Talent, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnLimboWaitHeal: func(mod *modifier.ModifierInstance) bool {
```
unleashurgeek, line 23: nit: lets move larger subscription functions to their own method rather than keeping anonymous like this

internal/character/gepard/talent.go
```diff
@@ -0,0 +1,74 @@
...
+				if mod.State().(talentState).a4Active {
+					mod.Engine().ModifyEnergyFixed(mod.Owner(), 100)
+				}
+
+				// If E6, action forward
+				if mod.State().(talentState).e6Active {
+					mod.Engine().SetGauge(mod.Owner(), 0)
+				}
+
+				mod.RemoveSelf()
```
unleashurgeek, line 51: these should be part of the `InsertAbility` Execute func (only happen if the revive goes through)

nit: fine with leaving the E6 as-is, but may need some changes. Need to follow up with some TC since it seems to have a weird edge case based on DM

internal/character/gepard/technique.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	targets := c.engine.Characters()
+
+	for _, trg := range targets {
+		c.engine.AddModifier(trg, info.Modifier{
+			Name:   Technique,
+			Source: c.id,
+		})
+	}
```
unleashurgeek, line 43: like with E4, need to handle case of `CharacterAdded` event

internal/character/gepard/trace.go
```diff
@@ -0,0 +1,43 @@
...
+const (
+	A2 key.Modifier = "gepard-a2"
+	A6 key.Modifier = "gepard-a6"
+)
+
+func init() {
+	modifier.Register(A6, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnPhase1: func(mod *modifier.ModifierInstance) {
+				mod.SetProperty(prop.ATKFlat, mod.Engine().Stats(mod.Owner()).DEF()*0.35)
```
unleashurgeek, line 19: You'll need to set the property `ATKConvert` instead of `ATKFlat`.

internal/character/gepard/ult.go
```diff
@@ -0,0 +1,55 @@
...
+type ultState struct {
+	shieldPerc float64
+	shieldFlat float64
+}
+
+func init() {
+	modifier.Register(Ult, modifier.Config{
+		Duration:   3,
+		StatusType: model.StatusType_STATUS_BUFF,
+		Listeners: modifier.Listeners{
```
unleashurgeek, line 24: Modifier should have `TickImmediately: true`

internal/character/gepard/technique.go
```diff
@@ -0,0 +1,44 @@
...
+
+const (
+	Technique       key.Modifier = "gepard-technique"
+	TechniqueShield key.Shield   = "gepard-technique-shield"
+)
+
+func init() {
+	modifier.Register(Technique, modifier.Config{
+		Duration:   2,
+		StatusType: model.StatusType_STATUS_BUFF,
```
unleashurgeek, line 18: modifier should have `TickImmediately: true`

internal/character/gepard/eidolon.go
Kisa75: This is not the resistance buff, but rather the tracker that the dms used to manage Gepards "alive"-status

internal/character/gepard/talent.go
```diff
@@ -20,39 +20,44 @@ type talentState struct {
...
+		Source:   mod.Owner(),
+		Priority: info.CharReviveSelf,
 	})
+
+	// If E6, action forward
+	if mod.State().(talentState).e6Active {
+		mod.Engine().SetGauge(mod.Owner(), 0)
+	}
+
+	mod.RemoveSelf()
```
unleashurgeek, line 59: This should also be moved inside the `Execute` insert (both action forward & `RemoveSelf`)

PR #108, opened 2023-06-29
Implemented missing OnBeforeDying modifier events (#19)

### LINKED ISSUE #19: Missing `OnBeforeDying` modifier event
Pending TC testing before we can commit to an implementation. Needs to be implemented for the following abilities (non-exhaustive):

- [ ] Pela A4

Kisa75:
Resolves #19

### DIFF
--- a/internal/character/silverwolf/eidolon.go
+++ b/internal/character/silverwolf/eidolon.go
@@ -24,6 +24,15 @@ func init() {
 			OnAdd: func(mod *modifier.ModifierInstance) {
 				mod.SetProperty(prop.EffectRES, -0.2)
 			},
+			OnBeforeDying: func(mod *modifier.ModifierInstance) {
+				if mod.Owner() == mod.Source() {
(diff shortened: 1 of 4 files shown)


internal/character/silverwolf/eidolon.go
```diff
@@ -24,6 +24,15 @@ func init() {
 			OnAdd: func(mod *modifier.ModifierInstance) {
 				mod.SetProperty(prop.EffectRES, -0.2)
 			},
+			OnBeforeDying: func(mod *modifier.ModifierInstance) {
+				if mod.Owner() == mod.Source() {
```
unleashurgeek, line 28: this solution won't work because this modifier is attached to enemies. My recommendation for here is in the eidolon init, create a subscription on the `TargetDeath` event and then do loop to remove from all enemies when SW dies.

unleashurgeek (approved):

PR #111, opened 2023-06-30
Implement Collapsing Sky [LC]

GortFan:


internal/lightcone/destruction/collapsingsky/collapsingsky.go
```diff
@@ -0,0 +1,63 @@
...
+	modifier.Register(Check, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeAttack: onBeforeAttack,
+			OnAfterAttack:  onAfterAttack,
+		},
+	})
+
+	modifier.Register(Buff, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 36: Looking at the datamine:
* This should just be 1 modifier that only subscribes to `OnBeforeHit` and modifies the damage data
* This modifier is permanently attached and is not registered as a `BUFF`

unleashurgeek (approved):

PR #120, opened 2023-07-03
Implemented Belobog of the Architects [Relic]

sky-larke:


internal/relic/planar/belobog/belobog.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func onPropChange(mod *modifier.Instance) {
+	stats := mod.OwnerStats()
+	if stats.EffectHitRate() >= 0.5 {
+		mod.SetProperty(prop.DEFPercent, 0.30)
+	} else {
+		mod.SetProperty(prop.DEFPercent, 0.15)
+	}
```
unleashurgeek, line 56: make this a single function that is registered as both listeners (they have the same signature)

unleashurgeek (approved):

PR #121, opened 2023-07-03
Implement Perfect Timing LC

RyanOld:


internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,60 @@
...
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	effResAmt := 0.12 + 0.04*float64(lc.Imposition)
+	engine.AddModifier(owner, info.Modifier{
+		Name:   PTEffRes,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectRES: effResAmt},
+		State:  float64(lc.Imposition), // 2 calcs outside depends on lc imp. level.
+	})
+}
```
unleashurgeek, line 44: Here is my suggestion on how to make this match the game behavior:
1. Remove all listeners from `PTEffRes` and the `State` (so this modifier now only adds `EffectRES` and does not manage healing boost)
2. Create a `giveHealBuff` method like you currently have, but make the signature `giveHealBuff(engine engine.Engine, owner *info.Stats, imposition int, prevBuff float64) float64`
  - this new method should calculate the heal buff amount like it currently does
  - if the calculated amount `!= prevBuff` that is passed in
    - call `AddModifier` and populate the `Stats` line with the new buff amount
    - return the new buff amount
  - if the calculated amount `== prevBuff`, do nothing and return the prevBuff amount (only add the modifier if the buff amount changes)
4. Create a `BattleStart` subscription that will call `giveHealBuff` on the owner target. Example of how to do this: https://github.com/simimpact/srsim/blob/6643c1b61df8bf8ba25064a8bbaa1a6b4f305b34/internal/relic/planar/vonwacq/vonwacq.go#L32-L38
5. Create a `ActionEnd` subscription that will call `engine.Stats` for the owner and then call `giveHealBuff`

A pseudo version of what this `Create` method might look like:
```go
func Create(eng engine.Engine, owner key.TargetID, lc info.LightCone) {
	effResAmt := 0.12 + 0.04*float64(lc.Imposition)
  healBuffAmt := 0

  // add the effect res buff
	engine.AddModifier(owner, info.Modifier{
		Name:   PTEffRes,
		Source: owner,
		Stats:  info.PropMap{prop.EffectRES: effResAmt},
	})

  engine.Events().BattleStart.Subscribe(func(e event.BattleStart) {
		for _, char := range e.CharStats {
			if char.ID() == owner {
				healBuffAmt = giveHealBuff(eng, char, lc.Imposition, healBuffAmt)
			}
		}
	})

  engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
    healBuffAmt = giveHealBuff(eng, eng.Stats(owner), lc.Imposition, healBuffAmt)
  })
}
```

internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,75 @@
...
+	modifier.Register(PTEffRes, modifier.Config{})
+
+	modifier.Register(PTHealBoost, modifier.Config{
+		Stacking: modifier.ReplaceBySource,
+	})
+}
+
+func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
+	effResAmt := 0.12 + 0.04*float64(lc.Imposition)
+	healBuffAmt := float64(0)
```
unleashurgeek, line 38: `healBuffAmt := 0.0`

internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		engine.AddModifier(owner.ID(), info.Modifier{
+			Name:   PTHealBoost,
+			Source: owner.ID(),
+			Stats:  info.PropMap{prop.HealBoost: healBuffAmt},
+		})
+	} else {
+		return prevBuff
+	}
+	return healBuffAmt
```
unleashurgeek, line 74: move this return inside their if, drop the else (just `return prevBuff` if the if condition fails)

internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,75 @@
...
+	})
+	// OnActionEnd -> recalc HealBoost.
+	engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
+		healBuffAmt = giveHealBuff(engine, engine.Stats(owner), lc.Imposition, healBuffAmt)
+	})
+}
+
+func giveHealBuff(engine engine.Engine, owner *info.Stats, imposition int, prevBuff float64) float64 {
+	currEffRes := engine.Stats(owner.ID()).EffectRES()
+	healBuffAmt := currEffRes*0.30 + 0.03*float64(imposition)
```
unleashurgeek, line 60: should be `currEffRes * (0.30 + 0.03*float64(imposition))`

internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,75 @@
...
+	engine.AddModifier(owner, info.Modifier{
+		Name:   PTEffRes,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectRES: effResAmt},
+	})
+	// OnBattleStart -> giveHealBoost to owner.
+	engine.Events().BattleStart.Subscribe(func(e event.BattleStart) {
+		for _, char := range e.CharStats {
+			if char.ID() == owner {
+				giveHealBuff(engine, char, lc.Imposition, healBuffAmt)
```
unleashurgeek, line 48: assign `healBuffAmt` here

internal/lightcone/abundance/perfecttiming/perfecttiming.go
```diff
@@ -0,0 +1,75 @@
...
+	// OnActionEnd -> recalc HealBoost.
+	engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
+		healBuffAmt = giveHealBuff(engine, engine.Stats(owner), lc.Imposition, healBuffAmt)
+	})
+}
+
+func giveHealBuff(engine engine.Engine, owner *info.Stats, imposition int, prevBuff float64) float64 {
+	currEffRes := engine.Stats(owner.ID()).EffectRES()
+	healBuffAmt := currEffRes*0.30 + 0.03*float64(imposition)
+	if healBuffAmt != prevBuff {
```
unleashurgeek, line 61: you should perform the max check **before** this check. In the current state, when past the max this check will always return true even when there is no change.

unleashurgeek (approved):

PR #124, opened 2023-07-04
Implemented Fleet [Relic]

sky-larke:
Not quite sure if I added modifier to all party members correctly, we don't have similar implementation for relics yet


internal/relic/planar/fleet/fleet.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/simimpact/srsim/pkg/engine/equip/relic"
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+)
+
+const (
+	mod = key.Modifier("fleet-of-the-ageless")
+)
```
unleashurgeek, line 14: mod name format

internal/relic/planar/fleet/fleet.go
```diff
@@ -0,0 +1,53 @@
...
+func init() {
+	relic.Register(key.FleetOfTheAgeless, relic.Config{
+		Effects: []relic.SetEffect{
+			{
+				MinCount: 2,
+				Stats:    info.PropMap{prop.HPPercent: 0.12},
+			},
+			{
+				MinCount: 2,
+				CreateEffect: func(engine engine.Engine, owner key.TargetID) {
```
unleashurgeek, line 27: nit: this create is more complex, so I would move the function definition outside init like `onCheck` to help make it easier to read and reduce depth

internal/relic/planar/fleet/fleet.go
```diff
@@ -0,0 +1,53 @@
...
+			OnPropertyChange: onCheck,
+		},
+	})
+}
+
+func onCheck(mod *modifier.Instance) {
+	stats := mod.OwnerStats()
+	if stats.SPD() >= 120 {
+		mod.SetProperty(prop.ATKPercent, 0.08)
+	}
```
unleashurgeek, line 52: this one is going to need to be a bit different. In the current state, you are adding this modifier onto all team members and then applying the ATK% buff depending on if the speed **of that team member** is greater than 120. Instead, the check should be done on the owner of this Relic where the applied modifier is rather simple. To this end, my suggestion:
1. You will need to create 2 modifiers, `fleet-of-the-ageless` (the speed check) and `fleet-of-the-ageless-atk-buff` (the ATK% increase) 
2. In the `CreateEffect` function have it do the following:
    1. subscribe to `BattleStart`. On battle start, . Example of how to subscribe: https://github.com/simimpact/srsim/blob/ffc47a1f1388b56051686508a453564c980059ba/internal/lightcone/harmony/chorus/chorus.go#L51-L55
3. Have the checker modifier have the `OnAdd` & `OnPropertyChange` listeners. This check should perform the speed check and apply/remove the buff modifier as needed
4. the check method should be written like the following (used by battlestart & prop change):
```go
func check(mod *modifier.Instance) {
  stats := mod.OwnerStats()
  applied := mod.State().(*bool)

  if stats.SPD() >= 120 && !applied {
    for _, c := range e.Characters() {
      mod.Engine().AddModifier(c, info.Modifier{
        Name:   buff,
        Source: owner,
        Stats:  info.PropMap{prop.ATKPercent, 0.08}
      })
    }
    mod.State().(*bool) = true
  }

  if stats.SPD() < 120 && applied {
    for _, c := range e.Characters() {
      mod.Engine().RemoveModifierFromSource(c, mod.Owner(), buff)
    }
    mod.State().(*bool) = false
  }
}
```
5. Check should have a state to keep track of if the attack buff is applied to reduce the amount of `AddModifier`/`RemoveModifier` calls it needs to perform
6. Buff modifier needs stacking behavior `ReplaceBySource`. If multiple allies have this set, you can have multiple ATK% buffs applied to each character (1 per set)

PR #131, opened 2023-07-07
Implement Incessant Rain (LC)

RyanOld:


internal/lightcone/nihility/incessantrain/incessantrain.go
unleashurgeek, line 1: The overall logic should be rewritten to better mimic game and reduce the number of modifiers used. Should only need two modifiers `incessant-rain` and `aether-code`

The `aether-code` modifier should be a simple 1 turn debuff similar to how `aetherCodeDebuff` is written now (just missing the stacking behavior)

The `incessant-rain` modifier should have the following:
* In the `Create`, add the EHR stat (like what is done for `ehrNCritCheck` now)
* `incessant-rain` should have the following listeners
  - `OnBeforeHitAll` for the crit rate hit buff (see other comment for fixes)
  - `OnAfterAttack` to capture list of attacked targets similar to how done in-game
  - `OnAfterAction` to perform applying aether code (they do this on action end rather than attack end for some reason, so need to mimic that behavior here which adds a bit more complexity)
* `incessant-rain` needs a state struct that contains the following (the state needs to also be passed as a pointer):
```go
type state struct {
  amt     float64
  targets []key.TargetID  // set to nil in Create
}
```
* In `OnAfterAttack`, all this listener should do is update `targets` in the state:
```go
func afterAttack(mod *modifier.Instance, e event.AttackEnd) {
  state := mod.State().(*state)
  state.targets = e.Targets
}
```
* In `OnAfterAction`, perform the apply aether code attempt. This should look somewhat like the following:
```go
func actionEnd(mod *modifier.Instance, e event.ActionEnd) {
  state := mod.State().(*state)

  // make new array for possible targets, loop through state.targets to filter
  validEnemies := make([]key.TargetID, 0, len(state.targets))
  for _, t := range state.targets {
    if mod.Engine().HPRatio(t) > 0 && !mod.Engine().HasModifier(t, code) {
      validEnemies = append(validEnemies, t)
    }
  }

  if validEnemyList != nil {
    // choose one enemy, apply debuff to them.
    chosenOne := validEnemyList[mod.Engine().Rand().Intn(len(validEnemyList))]
    mod.Engine().AddModifier(chosenOne, info.Modifier{
      Name:     code,
      Source:   mod.Owner(),
      Stats:    info.PropMap{prop.AllDamageTaken: state.amt},
      Chance:   1.0,
      Duration: 1,
    })
  }
  
  // set targets to nil at end to reset
  state.targets = nil
}
```

internal/lightcone/nihility/incessantrain/incessantrain.go
```diff
@@ -0,0 +1,107 @@
...
+	modState := state{
+		critAmt:     0.10 + 0.02*float64(lc.Imposition),
+		dmgTakenAmt: 0.10 + 0.02*float64(lc.Imposition),
+		targets:     nil,
+	}
+	engine.AddModifier(owner, info.Modifier{
+		Name:   rain,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectHitRate: ehrAmt},
+		State:  modState,
```
unleashurgeek, line 63: needs to be passed as a pointer, so `&modState`

internal/lightcone/nihility/incessantrain/incessantrain.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+}
+
+// boost CR on current hit if enemy has >=3 debuffs
+func critRateBoost(mod *modifier.Instance, e event.HitStart) {
+	state := mod.State().(*state)
+	debuffCount := float64(e.Hit.Defender.StatusCount(model.StatusType_STATUS_DEBUFF))
+	if debuffCount >= 3 {
+		e.Hit.Defender.AddProperty(prop.CritChance, state.critAmt)
```
unleashurgeek, line 72: should be buffing the attacker, not defender

unleashurgeek (approved):

PR #133, opened 2023-07-08
Clara implementation

mnpqraven:


internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/simimpact/srsim/pkg/engine/event"
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	E2Buff key.Modifier = "clara-e2-buff"
```
unleashurgeek, line 13: rename to just `clara-e2`

internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+
+// After using the Ultimate, ATK increases by 30% for 2 turn(s).
+func (c *char) e2() {
+	if c.info.Eidolon >= 2 {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:     E2Buff,
+			Source:   c.id,
+			Stats:    info.PropMap{prop.ATKPercent: 0.3},
+			Duration: 2,
+		})
```
unleashurgeek, line 66: Needs `TickImmediately: true` based on DM

internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+		Stacking:   modifier.Refresh,
+		StatusType: model.StatusType_STATUS_BUFF,
+		Listeners: modifier.Listeners{
+			OnAfterBeingAttacked: func(mod *modifier.Instance, e event.AttackEnd) {
+				mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+					Name:     E4Buff,
+					Source:   mod.Owner(),
+					Duration: 1,
+					Stats:    info.PropMap{prop.AllDamageReduce: 0.3},
+				})
```
unleashurgeek, line 39: `TickImmediately: true`

internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+
+func init() {
+	modifier.Register(E2Buff, modifier.Config{
+		Stacking:   modifier.Refresh,
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
+
+	modifier.Register(E4Buff, modifier.Config{
+		TickMoment: modifier.ModifierPhase1End,
+	})
```
unleashurgeek, line 27: StatusType `BUFF`

internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+
+	modifier.Register(E4, modifier.Config{
+		Stacking:   modifier.Refresh,
+		StatusType: model.StatusType_STATUS_BUFF,
```
unleashurgeek, line 31: Remove `StatusType` and `Stacking` from this modifier (it is logic, not a buff)

internal/character/clara/skill.go
```diff
@@ -0,0 +1,26 @@
...
+			model.DamageFormula_BY_ATK: skill[c.info.SkillLevelIndex()],
+		},
+		StanceDamage: 30.0,
+		EnergyGain:   30,
+	})
+
+	// E1 check
+	c.e1()
+
+	state.EndAttack()
```
unleashurgeek, line 25: `EndAttack` happens before E1 check

internal/character/clara/skill.go
```diff
@@ -0,0 +1,26 @@
...
+		Source:     c.id,
+		Targets:    c.engine.Enemies(),
+		DamageType: model.DamageType_PHYSICAL,
+		AttackType: model.AttackType_SKILL,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: skill[c.info.SkillLevelIndex()],
+		},
+		StanceDamage: 30.0,
+		EnergyGain:   30,
+	})
```
unleashurgeek, line 20: missing revenge hits. Full skill logic is as follows:
1. Hit all enemies that have counter mark (hit `AttackType == SKILL`, does 0 stance damage and gens 0 energy)
2. Hit all enemies (the hit you currently have defined)
3. `EndAttack`
4. Remove all marks if not E1

internal/character/clara/skill.go
```diff
@@ -0,0 +1,26 @@
...
+
+	// E1 check
+	c.e1()
```
unleashurgeek, line 23: nit: would rather inline the logic here so it is more clear that E0 is counter mark removal

internal/character/clara/technique.go
```diff
@@ -0,0 +1,20 @@
+package clara
+
+import (
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+)
+
+const (
+	Technique key.Modifier = "clara-technique-aggro"
```
unleashurgeek, line 10: Modifier needs to registered with `StatusType == BUFF`

internal/character/clara/trace.go
```diff
@@ -0,0 +1,64 @@
...
+						Count:  1,
+					})
+				}
+			},
+		},
+	})
+}
+
+func (c *char) initTraces() {
+	if c.info.Traces["1107101"] {
```
unleashurgeek, line 42: with https://github.com/simimpact/srsim/pull/134 merged, regenerate `data.go` and change trace ids to the new normalized value

internal/character/clara/trace.go
```diff
@@ -0,0 +1,64 @@
...
+		})
+	}
+
+	if c.info.Traces["1107103"] {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   A6,
+			Source: c.id,
+			Stats:  info.PropMap{prop.AllDamagePercent: 0.3},
+		})
+	}
```
unleashurgeek, line 63: This is incorrect. In current implementation, this will buff **all** Clara attacks by 30%

datamine implementation does the following:
1. modifier implemented with `OnBeforeHit` listener. Listener modifies hit data to give attacker 30% `AllDamagePercent` increase
2. modifier is added to clara at the start of counter
3. modifier is removed from clara at the end of counter

I think we can simplify the implementation to the following:
1. modifier implements `OnBeforeHit` listener
2. In `OnBeforeHit` listener, modify hit `AllDamagePercent` iff `AttackType == INSERT`
3. Apply this modifier on init (always on like the other two)

This variant should work and be equivalent since the only attacks by Clara that are `AttackType == INSERT` are her counter attacks

internal/character/clara/ult.go
```diff
@@ -0,0 +1,79 @@
...
+// immediately launches a Counter, and its DMG multiplier against the enemy
+// increases by *%. Enemies adjacent to it take 50% of the DMG dealt to the
+// target enemy. Enhanced Counter(s) can take effect 2 time(s).
+func (c *char) Ult(target key.TargetID, state info.ActionState) {
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   Ult,
+		Source: c.id,
+	})
+
+	c.e2()
```
unleashurgeek, line 76: nit: E2 before ult modifier (though I dont think it matters in this case)

internal/character/clara/ult.go
```diff
@@ -0,0 +1,79 @@
...
+				if mod.Engine().HasModifier(mod.Owner(), E6) {
+					amt = 3
+				}
+
+				mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+					Name:   UltCounter,
+					Source: mod.Owner(),
+					State:  ultCounterNum(amt),
+				})
+			},
```
unleashurgeek, line 43: * You are not setting state in `AddModifier` call so this logic fails
* Nothing in this logic seems to necessitate a `State` + `OnAdd` listener. Should be able to just set `Stats` and put this logic in the `Ult` function directly (so 1. create `clara-ult-buff`, 2. create `enhanced-counter` buff)
* When doing this directly in ult, you do not need to do the expensive `E6` modifier check and can just check her `eidolon` level directly
* `clara-ult-buff` needs to be registered as `StatusType == BUFF`

internal/character/clara/eidolon.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// After other allies are hit, Svarog also has a 50% fixed chance to trigger a
+	// Counter on the attacker and mark them with a Mark of Counter. When using
+	// Ultimate, the number of Enhanced Counters increases by 1.
+	if c.info.Eidolon >= 6 {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   E6,
+			Source: c.id,
+		})
+	}
```
unleashurgeek, line 88: Get rid of this modifier (see comment in `ult.go`)

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, target := range e.Targets {
+		if target == c.id {
+			isClara = true
+		}
+	}
+	if isClara || (c.engine.HasModifier(c.id, E6) && c.engine.Rand().Float32() < 0.5) {
+		return true
+	}
+	return false
+}
```
unleashurgeek, line 153: 1. You need to check if the non-clara targets are an ally
2. the E6 50% chance check needs to happen **per ally hit** (so if 3 allies are hit, that is 50% fixed change 3 times)
3. Missing check for if ult is active/has count (**note:** ult does not decrement until insert is executed)
4. Check clara's eidolon level rather than using E6 modifier


Simplified version:
```go
func (c *char) canCounter(e event.AttackEnd) bool {
  ult := c.engine.HasModifier(c.id, UltCounter)
  for _, target := range e.Targets {
    if !c.engine.IsCharacter(target) {
      continue  
    }

    if target == c.id || ult || (c.info.Eidolon >= 6 && c.engine.Rand().Float64() < 0.5) {
      return true
    }
  }
  return false
}
```

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+	attackerID := e.Attacker
+
+	// won't counter non-enemies (dominated ally)
+	if !c.engine.IsEnemy(attackerID) {
+		return
+	}
+
+	if c.canCounter(e) {
+		c.doCounter(attackerID)
+	}
```
unleashurgeek, line 75: The counter marker should be added prior to the `canCounter` check. The cases where the marker is added:
1. When clara gets hit
2. When E6 && an ally gets hit. **Note:** this always happens and is done prior to the 50% fixed change

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+			enhancedDamage := info.DamageMap{
+				model.DamageFormula_BY_ATK: talent[c.info.TalentLevelIndex()] +
+					ultDmgBoost[c.info.UltLevelIndex()],
+			}
+			mainTargetDamage := normalDamage
+
+			if c.engine.HasModifier(c.id, UltCounter) {
+				mainTargetDamage = enhancedDamage
+			}
```
unleashurgeek, line 95: change this to be based on the percent amount rather than whole damage map. Will simplify logic:

```go
ult := c.engine.HasModifier(c.id, UltCounter)
pct := talent[c.info.TalentLevelIndex()]
if ult {
  pct += ultDmgBoost[c.info.UltLevelIndex()]
}

c.engine.Attack(info.Attack{
  Source:       c.id,
  Targets:      []key.TargetID{attackerID},
  DamageType:   model.DamageType_PHYSICAL,
  AttackType:   model.AttackType_INSERT,
  BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: pct},
  StanceDamage: 30.0,
  EnergyGain:   5.0,
})

if ult {
  c.engine.Attack(info.Attack{
    Source:       c.id,
    Targets:      c.engine.AdjacentTo(attackerID),
    DamageType:   model.DamageType_PHYSICAL,
    AttackType:   model.AttackType_INSERT,
    BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: pct * 0.5},
    StanceDamage: 30.0,
  })
  
  // TODO: decrement ult counter
}
```

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+
+				c.engine.Attack(info.Attack{
+					Source:       c.id,
+					Targets:      c.engine.AdjacentTo(attackerID),
+					DamageType:   model.DamageType_PHYSICAL,
+					AttackType:   model.AttackType_INSERT,
+					BaseDamage:   splashDamage,
+					StanceDamage: 30.0,
+				})
+			}
```
unleashurgeek, line 122: Need to decrement the counter here and let the engine self-manage the modifier and delete it when count reaches 0

internal/character/clara/ult.go
```diff
@@ -0,0 +1,79 @@
...
+				if e.AttackType == model.AttackType_INSERT {
+					remaining := mod.State().(ultCounterNum)
+					remaining -= 1
+					if remaining <= 0 { // executes 2 counters
+						mod.RemoveSelf()
+					}
+				}
+			},
+		},
+	})
```
unleashurgeek, line 61: This is very wrong. This will have UltCounter reach 0 after first ult insert (since 2-hit attack, will have remaining reach 0).

This modifier should have no listener, just use the modifier `Count` system instead. This config then should only specify a max count of 3 (since `MaxCount` is required when using count). When adding the `UltCounter` modifier, set the starting count to 2 or 3 depending on E6 status

Modifier also needs stacking behavior `ReplaceBySource` so count gets reset if it is non-zero when you ult

internal/character/clara/talent.go
```diff
@@ -100,54 +101,53 @@ func (c *char) doCounter(attackerID key.TargetID) {
...
 
-// helper fn to see if a counter is eligible
-//
-// - if clara is not targeted AND < E6 -> no counter
-//
-// - clara targeted OR E6 + winning E6 50/50 -> counter
 func (c *char) canCounter(e event.AttackEnd) bool {
-	isClara := false
+	// clara has ult counter stacks left
+	hasUlt := c.engine.HasModifier(c.id, UltCounter) && c.engine.ModifierStackCount(c.id, c.id, UltCounter) > 0
```
unleashurgeek, line 134: only need one of these checks. Modifiers will be automatically removed when their count reaches 0. If you call `ModifierStackCount`, will return `0` iff you do not have the modifier. Current state is looping over Clara modifiers twice

internal/character/clara/ult.go
```diff
@@ -0,0 +1,73 @@
...
+	modifier.Register(Ult, modifier.Config{
+		BehaviorFlags: []model.BehaviorFlag{model.BehaviorFlag_STAT_BURST},
+		Stacking:      modifier.Refresh,
+		Duration:      2,
+	})
+
+	modifier.Register(UltBuff, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+		Duration:   2,
+	})
```
unleashurgeek, line 27: Merge these into a single buff. No reason to separate if their signature is the same (they are also a single modifier in the datamine)

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	// canCounter (clara targeted or e6 + an ally winning 50/50)
+	if c.canCounter(e) {
+		// add marker modifier on enemy attacker
+		c.engine.AddModifier(attackerID, info.Modifier{
+			Name:   TalentMark,
+			Source: c.id,
+			State:  State{skillLevelIndex: c.info.SkillLevelIndex(), ultLevelIndex: c.info.UltLevelIndex()},
+		})
```
unleashurgeek, line 80: * This is not 100% correct. Based on the datamine, the talent mark is added **before** the E6 50% fixed chance check (so mark is added even if that check fails)
* Why does the `TalentMark` have a state? Do not see it used anywhere

internal/character/clara/talent.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+
+			// remove marker modifier on all enemies when clara dies
+			OnBeforeDying: func(mod *modifier.Instance) {
+				for _, enemyID := range mod.Engine().Enemies() {
+					mod.Engine().RemoveModifier(enemyID, TalentMark)
+				}
+			},
+		},
+	})
```
unleashurgeek, line 46: I don't think `TalentCounter` modifier is necessary:
* The `TalentMark` can be added via `talentActionEndListener` for the Clara hit case
* Since the `TalentMark` has no logic or stats, we don't need to remove it if Clara dies

unleashurgeek (approved):

PR #146, opened 2023-07-10
Natasha implementation

aaron-hwang:


unleashurgeek:
Need to fix the merge. Seem to be deleting a bunch of files that should not be deleted as part of this PR.

internal/character/natasha/attack.go
```diff
@@ -0,0 +1,30 @@
+package natasha
+
+import (
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+var attackHits = []float64{1.0}
```
unleashurgeek, line 9: not necessary since single hit (can also omit the `HitRatio` field, will default to 1.0)

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+			Targets:    []key.TargetID{target},
+			Source:     c.id,
+			DamageType: model.DamageType_PHYSICAL,
+			AttackType: model.AttackType_PURSUED,
+			BaseDamage: info.DamageMap{
+				model.DamageFormula_BY_MAX_HP: 0.4,
+			},
+			StanceDamage: 0.0,
+			EnergyGain:   0.0,
+			HitRatio:     1.0,
```
unleashurgeek, line 149: Can leave these unspecified, these values you set are the defaults

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Create an extra attack instance of type pursued, only called by nat basic attack
+func (c *char) e6(target key.TargetID) {
+	if c.info.Eidolon >= 6 {
+		c.engine.Attack(info.Attack{
+			Targets:    []key.TargetID{target},
+			Source:     c.id,
+			DamageType: model.DamageType_PHYSICAL,
+			AttackType: model.AttackType_PURSUED,
```
unleashurgeek, line 143: apparently the attack type of this is actually `Normal`

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+	alreadyUsed         bool
+}
+
+func init() {
+
+	//Self heal if hp lower than 30% after gettting hit : Eidolon 1
+	modifier.Register(E1, modifier.Config{
+		//Refactor into own method maybe?
+		Listeners: modifier.Listeners{
+			OnAfterBeingAttacked: func(mod *modifier.Instance, e event.AttackEnd) {
```
unleashurgeek, line 36: for larger functions like this, create a separate function to reference rather than using an anonymous function for improved readability

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+	E2FlatHOT             float64      = 160
+	E4                    key.Modifier = "natasha-e4"
+)
+
+type E1State struct {
+	PercentageThreshold float64
+	HealScale           float64
+	HealFlat            float64
+	alreadyUsed         bool
+}
```
unleashurgeek, line 28: * Why use a state here? All of these values appear to be static, and you have defined them as constants. What value is a state providing?
* rather than have an already used state, just remove the E1 modifier after the heal goes off (like what datamine does). This solution is also nice because then in logs the `natasha-e1-autoheal` modifier only shows up while the heal is available

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+									HealValue: selfHealState.HealFlat,
+								})
+							},
+							Source: selfHealer,
+							AbortFlags: []model.BehaviorFlag{
+								model.BehaviorFlag_STAT_CTRL,
+								model.BehaviorFlag_DISABLE_ACTION},
+							Priority: info.CharHealSelf,
+						})
+						selfHealState.alreadyUsed = true
```
unleashurgeek, line 59: note for future: If you were going to use state for this, this will not work. To modify state, you need it passed in as a pointer. You are currently passing it in by copy so the modification is meaningless

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+func (c *char) e2(targets []key.TargetID) {
+	if c.info.Eidolon >= 2 {
+		for _, trg := range targets {
+			targetQualifies := c.engine.Stats(trg).CurrentHPRatio() <= E2ThresholdPercentage
+			if targetQualifies {
+				c.engine.AddModifier(trg, info.Modifier{
+					Name:     E2,
+					Source:   c.id,
+					Duration: 1,
+				})
```
unleashurgeek, line 130: Needs to be added with `TickImmediately: true`

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+						})
+						selfHealState.alreadyUsed = true
+					}
+				}
+			},
+		},
+	})
+
+	//Register E2
+	modifier.Register(E2, modifier.Config{
```
unleashurgeek, line 67: * `TickMoment` is `ModifierPhase1End`
* `StatusType` is `BUFF`

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,63 @@
...
+		Listeners: modifier.Listeners{
+			OnBeforeBeingHeal: func(mod *modifier.Instance, e *event.HealStart) {
+				mod.Engine().DispelStatus(e.Target.ID(), info.Dispel{
+					Status: model.StatusType_STATUS_DEBUFF,
+					Order:  model.DispelOrder_LAST_ADDED,
+					Count:  1,
+				})
+			},
+		},
+	})
```
unleashurgeek, line 31: This is incorrect. Current implementation would have Natasha get a debuff dispelled from herself every time she gets healed.

The A2 should not be a modifier. Like DM, want to just perform dispel in the skill logic

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,63 @@
...
+
+	//A4 outgoing heal buff
+	modifier.Register(A4, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
```
unleashurgeek, line 36: This is not registered as a buff (needs no config settings)

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,63 @@
...
+
+}
+
+func (c *char) initTraces() {
+
+	//A2
+	if c.info.Traces["1101101"] {
```
unleashurgeek, line 43: make sure to update trace ids to match new data

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,63 @@
...
+			Source: c.id,
+		})
+	}
+
+	//A4
+	if c.info.Traces["1101102"] {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   A4,
+			Source: c.id,
+			State:  info.PropMap{prop.HealBoost: 0.1},
```
unleashurgeek, line 55: This should be `Stats`, not `State`

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,63 @@
...
+		})
+	}
+
+	//A6
+	if c.info.Traces["1101103"] {
+		c.engine.ExtendModifierDuration(c.id, Skill, 1)
+	}
```
unleashurgeek, line 62: This does nothing. The trace check should just be part of the skill logic and increases the set duration of the skill heal modifier by 1

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,68 @@
...
+				OnPhase1: func(mod *modifier.Instance) {
+					char, _ := mod.Engine().CharacterInfo(mod.Source())
+					mod.Engine().Heal(info.Heal{
+						Source:    mod.Source(),
+						Targets:   []key.TargetID{mod.Owner()},
+						BaseHeal:  info.HealMap{model.HealFormula_BY_HEALER_MAX_HP: skillContinuous[char.SkillLevelIndex()]},
+						HealValue: skillContinuousFlat[char.SkillLevelIndex()],
+					},
+					)
+				},
```
unleashurgeek, line 34: Move the `OnPhase1` listener to its own function rather than keeping anonymous

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (c *char) Skill(target key.TargetID, state info.ActionState) {
+
+	healTarget := []key.TargetID{target}
+
+	natasha := c.id
+
+	healPercentage := info.HealMap{
+		model.HealFormula_BY_HEALER_MAX_HP: skill[c.info.SkillLevelIndex()],
+	}
```
unleashurgeek, line 48: what is the value of defining these variables here rather than in the `info.Heal` struct directly (and just put the struct creation directly inside the heal call)?

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,68 @@
...
+						HealValue: skillContinuousFlat[char.SkillLevelIndex()],
+					},
+					)
+				},
+			},
+		},
+	)
+}
+
+func (c *char) Skill(target key.TargetID, state info.ActionState) {
```
unleashurgeek, line 40: Skill logic should be the following in order:
1. If A2 is enabled, dispel debuff on passed in target
2. Perform heal
3. Add modifier (where duration is 2 or 3 depending on if A6 is enabled)
4. Add 30 energy to Natasha

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,68 @@
...
+const (
+	Skill key.Modifier = "natasha-skill"
+)
+
+func init() {
+
+	//Nat HOT from skill
+	modifier.Register(
+		Skill,
+		modifier.Config{
```
unleashurgeek, line 19: Needs `TickMoment: ModifierPhase1End`

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,68 @@
...
+			Duration:          2,
+			Stacking:          modifier.Unique,
```
unleashurgeek, line 21: Stacking behavior is `ReplaceBySource` (want new applications to reset the duration/timer)

internal/character/natasha/talent.go
```diff
@@ -0,0 +1,39 @@
...
+func init() {
+	modifier.Register(Talent,
+		modifier.Config{
+			StatusType:        model.StatusType_STATUS_BUFF,
+			CanModifySnapshot: true,
+			Listeners: modifier.Listeners{
+				OnBeforeDealHeal: func(mod *modifier.Instance, e *event.HealStart) {
+					char, _ := mod.Engine().CharacterInfo(mod.Owner())
+					if e.Target.CurrentHPRatio() <= talentHpThresholdPercentage {
+						mod.AddProperty(prop.HealBoost, talent[char.TalentLevelIndex()])
```
unleashurgeek, line 26: Should **not** be adding the modifier, but modifying the heal data instead. Calling `mod.AddProperty` here will not modify this heal (instead gives Nat a permanent heal buff for her next heal).

internal/character/natasha/technique.go
```diff
@@ -0,0 +1,10 @@
+package natasha
+
+import (
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/key"
+)
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	
+}
```
unleashurgeek, line 10: missing implementation

internal/character/natasha/talent.go
```diff
@@ -0,0 +1,39 @@
...
+	talentHpThresholdPercentage float64      = 0.3
+)
+
+func init() {
+	modifier.Register(Talent,
+		modifier.Config{
+			StatusType:        model.StatusType_STATUS_BUFF,
+			CanModifySnapshot: true,
+			Listeners: modifier.Listeners{
+				OnBeforeDealHeal: func(mod *modifier.Instance, e *event.HealStart) {
```
unleashurgeek, line 23: move this to a qualified function

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+		})
+	}
+}
+
+// Add a HOT if heal target is 30% hp or lower when healed
+// Should only be called by Nat's ult
+func (c *char) e2(targets []key.TargetID) {
+	if c.info.Eidolon >= 2 {
+		for _, trg := range targets {
+			targetQualifies := c.engine.Stats(trg).CurrentHPRatio() <= E2ThresholdPercentage
```
unleashurgeek, line 124: dont use `Stats(trg).CurrentHPRatio`, use `engine.HPRatio(trg)` instead

internal/character/natasha/eidolon.go
```diff
@@ -0,0 +1,153 @@
...
+
+	//Self heal if hp lower than 30% after gettting hit : Eidolon 1
+	modifier.Register(E1, modifier.Config{
+		//Refactor into own method maybe?
+		Listeners: modifier.Listeners{
+			OnAfterBeingAttacked: func(mod *modifier.Instance, e event.AttackEnd) {
+				selfHealer := mod.Owner()
+				selfHealState := mod.State().(E1State)
+				if !selfHealState.alreadyUsed {
+					lowEnough := mod.OwnerStats().CurrentHPRatio() <= selfHealState.PercentageThreshold
```
unleashurgeek, line 40: dont use `OwnerStats().CurrentHPRatio()`, use `mod.Engine().HPRatio(mod.Owner())` instead

internal/character/natasha/ult.go
```diff
@@ -0,0 +1,22 @@
+package natasha
+
+import (
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+func (c *char) Ult(target key.TargetID, state info.ActionState) {
```
unleashurgeek, line 9: Ult should do the following in order:
1. Apply E2 to any valid characters
2. Perform party-wide heal
3. Add 5 energy to natasha

internal/character/natasha/ult.go
```diff
@@ -0,0 +1,22 @@
...
+	targets := c.engine.Characters()
+	c.engine.Heal(info.Heal{
+		Targets: targets,
+		Source:  c.id,
+		BaseHeal: info.HealMap{
+			model.HealFormula_BY_HEALER_MAX_HP: ult[c.info.UltLevelIndex()],
+		},
+	})
```
unleashurgeek, line 17: Missing the flat healing to add to this

aaron-hwang:
Meant to separate fixes into their own commits but accidentally bundled them all together: They should all be in there

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,81 @@
...
+		Source:  c.id,
+		BaseHeal: info.HealMap{
+			model.HealFormula_BY_HEALER_MAX_HP: skill[c.info.SkillLevelIndex()],
+		},
+		HealValue:   skillFlatHeal[c.info.SkillLevelIndex()],
+		UseSnapshot: true,
+	}
+
+	// The actual act of healing
+	c.engine.Heal(heal)
```
unleashurgeek, line 53: nit: merge these into a single line:
```go
c.engine.Heal(info.Heal{
  	Targets: []key.TargetID{target},
		Source:  c.id,
		BaseHeal: info.HealMap{
			model.HealFormula_BY_HEALER_MAX_HP: skill[c.info.SkillLevelIndex()],
		},
		HealValue:   skillFlatHeal[c.info.SkillLevelIndex()],
		UseSnapshot: true,
})
```

internal/character/natasha/skill.go
```diff
@@ -0,0 +1,81 @@
...
+
+	hotDuration := 2
+
+	// A6
+	if c.info.Traces["1101103"] {
+		hotDuration = 3
+	}
```
unleashurgeek, line 60: bring these lines closer together and switch to normalized trace id:
```go
// A6
hotDuration := 2
if c.info.Traces["103"] {
  hotDuration = 3
}
```

internal/character/natasha/talent.go
```diff
@@ -0,0 +1,42 @@
...
+
+const (
+	Talent                      key.Modifier = "natasha-talent"
+	talentHpThresholdPercentage float64      = 0.3
+)
+
+func init() {
+	modifier.Register(Talent,
+		modifier.Config{
+			StatusType:        model.StatusType_STATUS_BUFF,
```
unleashurgeek, line 20: This modifier is not type `BUFF`

internal/character/natasha/technique.go
```diff
@@ -0,0 +1,43 @@
...
+	targets := c.engine.Enemies()
+	c.engine.Attack(info.Attack{
+		Targets:    targets,
+		Source:     c.id,
+		AttackType: model.AttackType_MAZE,
+		DamageType: model.DamageType_PHYSICAL,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: techniqueScalePercentage,
+		},
+	})
```
unleashurgeek, line 34: * Need to call `state.EndAttack()` after this `Attack` call and before the add modifier calls to match dm 
* The attack is supposed to be against a random enemy. This current logic is performing an AoE attack against **all** enemies

internal/character/natasha/technique.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, trg := range targets {
+		c.engine.AddModifier(trg, info.Modifier{
+			Name:   technique,
+			Source: c.id,
+			Chance: 1,
+			Stats:  info.PropMap{prop.AllDamageReduce: 0.30},
```
unleashurgeek, line 40: `AllDamageReduce` means that all enemies will take 30% less damage. Instead, the prop you want to set is `Fatigue` (same prop used in the datamine): `info.PropMap{prop.Fatigue: 0.30}`

internal/character/natasha/technique.go
```diff
@@ -0,0 +1,43 @@
...
+	technique                key.Modifier = "natasha-technique"
+	techniqueScalePercentage float64      = 0.80
+)
+
+func init() {
+	modifier.Register(technique, modifier.Config{
+		StatusType: model.StatusType_STATUS_DEBUFF,
+		Duration:   1,
+		TickMoment: modifier.ModifierPhase1End,
+	})
```
unleashurgeek, line 21: * Needs BehaviorFlag `STAT_FATIGUE`
* Stacking: `ReplaceBySource`
* `TickMoment` should be unset/left at default of `ModifierPhase2End`

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,42 @@
...
+		Listeners: modifier.Listeners{
+			OnBeforeBeingHeal: func(mod *modifier.Instance, e *event.HealStart) {
+				mod.Engine().DispelStatus(e.Target.ID(), info.Dispel{
+					Status: model.StatusType_STATUS_DEBUFF,
+					Order:  model.DispelOrder_LAST_ADDED,
+					Count:  1,
+				})
+			},
+		},
+	})
```
unleashurgeek, line 30: This modifier (and `A2` name variable) is unused and should be deleted

internal/character/natasha/trace.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	A2 key.Modifier = "natasha-a2"
+	A4 key.Modifier = "natasha-a4"
+	A6 key.Modifier = "natasha-a6"
```
unleashurgeek, line 15: `A6` modifier key is not used and should be deleted

unleashurgeek (approved):

PR #149, opened 2023-07-12
Testing Implemented for `Engine.Retarget()`

RyanOld:
Changes :
- Implement Unit Testing (might miss some edge cases)
- Fixed .Max bug
- Filter function is now optional


pkg/simulation/target.go
```diff
@@ -94,16 +94,19 @@ func (sim *Simulation) Neutrals() []key.TargetID {
...
+		i := 0
+		for _, target := range data.Targets {
+			if (data.IncludeLimbo || sim.Attr.HPRatio(target) > 0) && data.Filter(target) {
+				data.Targets[i] = target
+				i++
+			}
 		}
+		// truncate to safely remove unqualified targets.
+		data.Targets = data.Targets[:i]
 	}
```
unleashurgeek, line 109: * We should still be filtering out limbo when `Filter` is `nil` (should be equivalent to having a filter that always returns `true`). 
* Add tests for having `Filter` unspecified and that it still filtering limbo is the expected behavior

pkg/simulation/target_test.go
```diff
@@ -23,7 +23,54 @@ func NewSim(t *testing.T, seed int64) (*simulation.Simulation, *mock.MockAttribu
 	return sim, attr
 }
 
-func TestRetargetSimple(t *testing.T) {
+// test var : Targets list
+func TestRetargetEmptyTargets(t *testing.T) {
+	sim, attr := NewSim(t, 1)
+
+	// for all targets, return full health in this test
+	attr.EXPECT().HPRatio(gomock.Any()).Return(1.0).AnyTimes()
```
unleashurgeek, line 31: Remove this `EXPECT`. For this test, we shouldn't expect `attr` to be called (since no input targets). By removing the `EXPECT`, the test will fail if for some reason we suddenly call `attr.HPRatio`

pkg/simulation/target_test.go
```diff
@@ -23,7 +23,54 @@ func NewSim(t *testing.T, seed int64) (*simulation.Simulation, *mock.MockAttribu
...
+
+	result := sim.Retarget(info.Retarget{
+		Targets: []key.TargetID{},
+		Filter:  func(target key.TargetID) bool { return target == 2 },
+		Max:     3,
+	})
+
+	assert.Equal(t, []key.TargetID{}, result)
```
unleashurgeek, line 39: Make sure you are using the right/best assert for your test. In this case, `assert.Empty(t, result)` makes more sense here

unleashurgeek (approved):

PR #152, opened 2023-07-14
Implement Time Waits for No One (LC)

RyanOld:


internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+		// after apply added dmg, set cd to 0
+		state.cooldown = 0
+	}
+}
+
+// take struct pointer, modify lastHealAmt value.
+func recordHealAmt(mod *modifier.Instance, e event.HealEnd) {
+	state := mod.State().(*healRecorder)
+	state.lastHealAmt = e.HealAmount
+}
```
unleashurgeek, line 113: This should be a sum of all healings that occur and not just the last heal amount. The amount resets after the attack occurs (so if you do 2 heals -> attack, get the sum of both heals as the `dmgAmt`.

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+
+// take struct pointer, modify cooldown value
+func refreshCD(mod *modifier.Instance) {
+	state := mod.State().(*healRecorder)
+	state.cooldown = 1
+}
+
+// if cooldown = 1, Retarget(1 target), add dmg type pursued, byPureDamage, ele same as holder
+func applyExtraDmg(mod *modifier.Instance, e event.AttackEnd) {
```
unleashurgeek, line 76: This needs to be a subscription and not a listener. As a listener, this will only run when the character it is attached to performs an attack. What you want is when any ally performs an attack. So this needs to:
1. subscribe to the `AttackEnd` event
2. short circuit if the attacker is not a character
3. get the modifier instance attached to the LC owner
4. short circuit if it is on CD
5. otherwise, perform attack, reset dmgAmt, and put mod on CD

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	time key.Modifier = "time-waits-for-no-one"
+	// key for extra dmg attack, not for modifier.
+	extraDmg key.Attack = "time-waits-for-no-one-extra-damage"
+)
+
+type healRecorder struct {
+	cooldown    int
```
unleashurgeek, line 21: cooldown as an `int` is not clear. Given how it is currently used, would recommend changing to a bool and having it make more sense in usage (only perform an attack if `cooldown == false`

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	time key.Modifier = "time-waits-for-no-one"
+	// key for extra dmg attack, not for modifier.
+	extraDmg key.Attack = "time-waits-for-no-one-extra-damage"
```
unleashurgeek, line 17: nit: I would use the same key for both modifier and attack, since context will make it clear if the key is being used in the modifier or the attack. Would change this to:
```go
name = "time-waits-for-no-one"
```

This makes the `name` variable an `untyped string` which will work as both `key.Modifier` and `key.Attack`

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+func applyExtraDmg(mod *modifier.Instance, e event.AttackEnd) {
+	state := mod.State().(*healRecorder)
+	dmgAmt := mod.State().(*healRecorder).lastHealAmt
+	if state.cooldown == 1 {
+		validTargets := e.Targets
+		chosenEnemy := mod.Engine().Retarget(info.Retarget{
+			// targets are enemies hit by this atk. NOTE : confirm if onLimbo is included.
+			Targets: validTargets,
+			Max:     1,
+		})
```
unleashurgeek, line 85: this is correct, the came has a similar `UseLimbo` flag which would be used if limbo'd targets are included

nit: I would get rid of `validTargets` variable and just pass in `e.Targets` directly

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+func refreshCD(mod *modifier.Instance) {
+	state := mod.State().(*healRecorder)
+	state.cooldown = 1
+}
+
+// if cooldown = 1, Retarget(1 target), add dmg type pursued, byPureDamage, ele same as holder
+func applyExtraDmg(mod *modifier.Instance, e event.AttackEnd) {
+	state := mod.State().(*healRecorder)
+	dmgAmt := mod.State().(*healRecorder).lastHealAmt
+	if state.cooldown == 1 {
```
unleashurgeek, line 79: flip condition and return to reduce code depth

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+			// NOTE : might need to later change BaseDmg field to optional later
+			BaseDamage: info.DamageMap{
+				model.DamageFormula_BY_ATK: 0.0,
+			},
+			// this attack shouldn't call onHit listeners right?
+			UseSnapshot: true,
+		})
+		// after apply added dmg, set cd to 0
+		state.cooldown = 0
+	}
```
unleashurgeek, line 106: nit: pass in `holderInfo.Element` directly rather than creating `dmgType` variable

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,113 @@
...
+// take struct pointer, modify cooldown value
+func refreshCD(mod *modifier.Instance) {
+	state := mod.State().(*healRecorder)
+	state.cooldown = 1
+}
+
+// if cooldown = 1, Retarget(1 target), add dmg type pursued, byPureDamage, ele same as holder
+func applyExtraDmg(mod *modifier.Instance, e event.AttackEnd) {
+	state := mod.State().(*healRecorder)
+	dmgAmt := mod.State().(*healRecorder).lastHealAmt
```
unleashurgeek, line 78: missing the damage multiplier. The damage is not total healing amount but `36%/42%/48%/54%/60%` of total heal amount

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// if onCooldown = 1, Retarget(1 target), add dmg type pursued, byPureDamage, ele same as holder
+func applyExtraDmg(engine engine.Engine, mod info.Modifier, e event.AttackEnd) {
+	state := mod.State.(*healRecorder)
+	dmgAmt := state.recordedHeals * state.extraDmgMult
+	// return early if on cd. reduces code depth
+	if state.onCooldown {
+		return
+	}
```
unleashurgeek, line 90: you are checking cooldown twice, once in the `AttackEnd` subscription and once here. You should only check once

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+}
+
+func refreshCD(mod *modifier.Instance) {
+	state := mod.State().(*healRecorder)
+	state.onCooldown = false
+}
+
+// if onCooldown = 1, Retarget(1 target), add dmg type pursued, byPureDamage, ele same as holder
+func applyExtraDmg(engine engine.Engine, mod info.Modifier, e event.AttackEnd) {
```
unleashurgeek, line 84: I would not take in the entire modifier but only what you need. I would change the signature to the following:
```go
func applyExtraDmg(engine engine.Engine, targets []key.TargetID, owner key.TargetID, state *healRecorder)
```

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
unleashurgeek, line 1: If you want to change to the subscription-only approach I would do the following:

1. Remove all listeners from the `time-waits-for-no-one` modifier (this modifier only gives the HP% and outgoing healing % increase and does nothing else)
2. Create a new `time-waits-for-no-one-cd` modifier with `TickMoment: ModifierPhase1End` and `Duration: 1`
3. change the `Create` function to instead do the following:
```go
func Create(engine engine.Engine, owner key.TargetID, lc info.LightCone) {
  engine.AddModifier(owner, info.Modifier{
    Name:   time,
    Source: owner,
    Stats: info.PropMap{
      prop.HPPercent: 0.15 + 0.03*float64(lc.Imposition),
      prop.HealBoost: 0.10 + 0.02*float64(lc.Imposition),
    },
  })

  extraDmgMult := 0.30 + 0.06*float64(lc.Imposition)
  healAmt := 0.0

  engine.Events().HealEnd.Subscribe(func(e event.HealEnd) {
    if e.Healer == owner {
      healAmt += e.HealAmount
    }
  })

  engine.Events().AttackEnd.Subscribe(func(e event.AttackEnd)) {
    if engine.IsCharacter(e.Attacker) && !engine.HasModifier(owner, cd) {
      applyExtraDmg(engine, e.Targets, owner, healAmt * extraDmgMult)

      healAmt = 0.0 // reset heal amount
      engine.AddModifier(owner, info.Modifier{ // add 1 turn cd modifier
        Name:     cd,
        Source:   owner,
        Duration: 1,
      })
    }
  }
}

func applyExtraDmg(engine engine.Engine, targets []key.TargetID, owner key.TargetID, amt float64) {
  // perform pursued attack in here
}
```

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,95 @@
...
+func init() {
+	lightcone.Register(key.TimeWaitsforNoOne, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        5,
+		Path:          model.Path_ABUNDANCE,
+		Promotions:    promotions,
+	})
+	modifier.Register(time, modifier.Config{})
+	// added to automatically track extraDmg cooldown
+	modifier.Register(cd, modifier.Config{})
```
unleashurgeek, line 35: CD needs to have its `TickMoment` as `ModifierPhase1End` so it disables at the start of the turn and not the end (which is what it currently does with the  default `ModifierPhase2End` tick moment)

internal/lightcone/abundance/timewaitsfornoone/timewaitsfornoone.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	time = "time-waits-for-no-one"
+	cd   = "time-waits-for-no-one-extra-dmg-cd"
```
unleashurgeek, line 16: nit: would just name this `time-waits-for-no-one-cd` since the dmg is the only thing that can go on cd

unleashurgeek (approved):

PR #153, opened 2023-07-14
Implement Before the Tutorial Mission Starts (SW Event LC)

RyanOld:


internal/lightcone/nihility/beforethetutorialmissionstarts/beforethetutorialmissionstarts.go
```diff
@@ -0,0 +1,65 @@
...
+	// Retarget : filter : has def down behavior flag, Max 1, includeLimbo
+	qualified := mod.Engine().Retarget(info.Retarget{
+		Targets: e.Targets,
+		Filter: func(target key.TargetID) bool {
+			// returns true if target has triggerFlags, otherwise false
+			return mod.Engine().HasBehaviorFlag(target, triggerFlags...)
+		},
+		Max:          1,
+		IncludeLimbo: true,
+	})
```
unleashurgeek, line 61: nit: I would change this to a helper function instead of using retarget. With retarget it'll have to call `HasBehaviorFlag` on all targets even if one already has the flag. Instead I'd write a helper function for this that has the ability to exit early:
```go
func hasDefReducedTarget(engine engine.Engine, targets []key.TargetID) bool {
  for _, target := range targets {
    if engine.HasBehaviorFlag(target, model.BehaviorFlag_STAT_DEF_DOWN) {
      return true
    }
  }
  return false
}
```

unleashurgeek (approved):

unleashurgeek (approved):

PR #154, opened 2023-07-14
Implement Shields [Core]

GortFan:


pkg/engine/shield/absorb.go
```diff
@@ -15,21 +15,26 @@ func (mgr *Manager) AbsorbDamage(target key.TargetID, damage float64) float64 {
...
-			ID:     id,
+			ID:     shield.name,
 			Target: target,
 			OldHP:  prevHP,
 			NewHP:  shield.HP,
 		})
 
 		if shield.HP == 0 {
-			mgr.RemoveShield(id, target)
+			mgr.RemoveShield(shield.name, target)
```
unleashurgeek, line 32: You cannot remove entries from a list while iterating over a list. This will cause the iteration to break and incorrectly iterate over the list. See how modifier remove defers all the removal calls to after the iteration is complete (and builds a separate list of entries to remove):

https://github.com/simimpact/srsim/blob/12711171278f35df30a278955b67435af4708556/pkg/engine/modifier/remove.go#L9-L22

pkg/engine/shield/absorb.go
```diff
@@ -15,21 +15,26 @@ func (mgr *Manager) AbsorbDamage(target key.TargetID, damage float64) float64 {
...
+			damage -= prevHP
+		}
+
+		if shield.HP != 0 {
+			damage = 0
 		}
```
unleashurgeek, line 38: This current logic is incorrect shield behavior. For some incoming damage X, that damage X is applied to **all** shields that are currently equipped (say you take 100 damage and have a 200 HP shield and a 150 HP shield. After the hit you will have a 100 HP shield and a 50 HP shield).

Current logic has first shield absorb damage, and the remaining goes to the next and so forth (current logic can also have `damage` go negative).

pkg/engine/shield/absorb.go
```diff
@@ -15,21 +15,26 @@ func (mgr *Manager) AbsorbDamage(target key.TargetID, damage float64) float64 {
...
 		if shield.HP < 0 {
 			shield.HP = 0
 		}
 		mgr.event.ShieldChange.Emit(event.ShieldChange{
-			ID:     id,
+			ID:     shield.name,
 			Target: target,
 			OldHP:  prevHP,
 			NewHP:  shield.HP,
 		})
```
unleashurgeek, line 29: We should only emit a `ShieldChange` event if the shield health changes. Before this loop, should short circuit and just return 0 if the incoming damage is 0

pkg/engine/shield/add.go
```diff
@@ -4,20 +4,46 @@ import (
 	"github.com/simimpact/srsim/pkg/engine/event"
 	"github.com/simimpact/srsim/pkg/engine/info"
 	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
 )
 
 func (mgr *Manager) AddShield(id key.Shield, shield info.Shield) {
 	// TODO: LOGIC FOR ADDING A SHIELD
 	// 1. Check if the target already has this shield, if so remove old (mgr.RemoveShield)
```
unleashurgeek, line 12: remove TODO comments and add your own

pkg/engine/shield/add.go
```diff
@@ -4,20 +4,46 @@ import (
...
+	if mgr.HasShield(shield.Target, id) {
+		mgr.RemoveShield(id, shield.Target)
+	}
```
unleashurgeek, line 15: nit: this is inefficient, you are looping over the shields list twice (once to find if you have the shield, 2nd to remove in the worst case). Better if we can reduce this to one loop.

Also I am not sure if the right approach is to emit the `RemoveShield` event or not. Based on the latest TC findings, we **do not** emit a `Remove` event in the case of modifiers being replaced, so it is probably best to do the same here

pkg/engine/shield/add.go
```diff
@@ -4,20 +4,46 @@ import (
...
+		case model.ShieldFormula_SHIELD_BY_SHIELDER_DEF:
+			shieldHP += v * mgr.attr.Stats(shield.Source).DEF()
+		case model.ShieldFormula_SHIELD_BY_SHIELDER_MAX_HP:
+			shieldHP += v * mgr.attr.Stats(shield.Source).HP()
+		case model.ShieldFormula_SHIELD_BY_TARGET_MAX_HP:
+			shieldHP += v * mgr.attr.Stats(shield.Target).HP()
+		case model.ShieldFormula_SHIELD_BY_SHIELDER_TOTAL_SHIELD:
+			shieldHP += shield.ShieldValue
+		}
+	}
```
unleashurgeek, line 34: * nit: Make the `Stats` calls before this loop. We want to minimize the amount of times we need to call `Stats` as much as possible (this will call `Stats` multiple times for the same target in the case of multiple formulas used)
* for `SHIELD_BY_SHIELDER_TOTAL_SHIELD` I believe this formula should be v times the current maximum shield HP attached to the shielder. Think of it like "duplicating" the shield from shielder to target. Should make helper to get the max/total shield HP

pkg/engine/shield/add.go
```diff
@@ -4,20 +4,46 @@ import (
...
+
+	newInstance := &Instance{name: id, HP: shieldHP}
```
unleashurgeek, line 36: missing the use of the `ShieldTaken` and `ShieldAbsorb` properties: https://github.com/simimpact/srsim/blob/12711171278f35df30a278955b67435af4708556/pkg/engine/prop/prop.go#L60-L64

These properties buff the shield HP being applied. These work similar to Heal: https://github.com/simimpact/srsim/blob/12711171278f35df30a278955b67435af4708556/pkg/engine/combat/heal.go#L52-L55

pkg/engine/shield/remove.go
```diff
@@ -11,7 +11,17 @@ func (mgr *Manager) RemoveShield(id key.Shield, target key.TargetID) {
...
+	i := 0
+	var removedShield []*Instance
+	for _, shield := range mgr.targets[target] {
+		if shield.name == id {
+			removedShield = append(removedShield, shield)
+		} else {
+			mgr.targets[target][i] = shield
+			i++
+		}
+	}
```
unleashurgeek, line 23: Delete `removedShield` since it is unused (should only be necessary if you are removing multiple shields at once).

pkg/engine/shield/absorb.go
```diff
@@ -6,32 +6,51 @@ import (
...
-		if shield.HP < 0 {
-			shield.HP = 0
+		// Getting outgoing damage
+		maxShield := mgr.MaxShield(target)
+		maxShieldHP := mgr.targets[target][maxShield]
+		newMaxShieldHP := maxShieldHP.hp - damage
+		damageOut := -newMaxShieldHP
+		if damageOut < 0 {
+			damageOut = 0
 		}
```
unleashurgeek, line 36: I'm not seeing how this works. Lets say I have a 50 HP shield equipped and the `AbsorbDamage` call passes in `100` damage:
1. Loop over all shields (just this one shield) and subtract `100` damage from its HP
2. shield is removed since its HP becomes 0 (`50 - 100 < 0`)
3. Get the max shield. Because I no longer have a shield equipped, this will return no shield (so assuming `maxShield.hp = 0`)
4. Do `maxShield.hp - damage`, which returns `-100` since I currently have no shield
5. `damageOut` set to `100` (the original incoming damage) and it is ignored that the shield I had absorbed 50 damage

My suggestions are the following:
1. move this logic to inside where you subtract shield hp so you only need to loop over shields once
2. Determine your `damageOut` and `maxShieldHP` in this loop by finding the min damage out and `maxShieldHP`

For the 2nd, you can do conditions like the following:
```go
maxShieldHP := 0.0
damageOut := damage
for _, shield := range mgr.targets[target] {
  // Dim returns the maximum of x-y or 0.
  remaining := math.Dim(damage, shield.hp)
  if remaining < damageOut {
    damageOut = remaining
  }

  shield.hp = math.Dim(shield.hp, damage)
  if shield.hp > maxShieldHP {
    maxShieldHP = shield.hp
  }

  // add to remove list or insert back in targets array
}
```

pkg/engine/shield/add.go
```diff
@@ -3,21 +3,65 @@ package shield
...
+
+	// Compute final shieldHP using shield HP formula
+	shieldHP := baseHP * (1 + source.GetProperty(prop.ShieldBoost)) * (1 + target.GetProperty(prop.ShieldTaken))
+
+	// Create new instance to add to list of shields for target
+	newInstance := &Instance{name: id, hp: shieldHP}
+
+	switch isMatching, index := mgr.CheckMatching(id, shield); isMatching {
+	// Replace shield at matching index
+	case isMatching == true:
```
unleashurgeek, line 44: to add to linter comment, the case can also be simplified to just:
```go
case true:
  //
case false:
  //
```
because you pass `isMatching` as the switch value/target

pkg/engine/shield/add.go
```diff
@@ -3,21 +3,65 @@ package shield
...
 }
+
+// Check list of shields for a matching shield
+func (mgr *Manager) CheckMatching(id key.Shield, shield info.Shield) (bool, int) {
+	for i, shields := range mgr.targets[shield.Target] {
+		if shields.name == id {
+			return true, i
+		}
+	}
+	return false, 0
```
unleashurgeek, line 66: nit: in this case have return be `return false, -1` so it will crash the program if someone tries to use it as an index

pkg/engine/shield/add.go
```diff
@@ -3,21 +3,65 @@ package shield
...
-	// 2. Get the stats for the source and target
-	// 3. Compute shield HP/create ShieldInstance given the add paramsm
-	// 4. add shield to mgr.targets[shield.target]
-	// 5. emit ShieldAdded event
-
-	// emit to signify shield added
+
+	// Get the stats for the source and target
+	source := mgr.attr.Stats(shield.Source)
+	maxShield := mgr.targets[shield.Target][mgr.MaxShield(shield.Target)]
```
unleashurgeek, line 15: this will cause the sim to crash if this target has no shields (should be 0 if they have no shields)

pkg/engine/shield/manager.go
```diff
@@ -41,3 +42,15 @@ func (mgr *Manager) IsShielded(target key.TargetID) bool {
...
+	maxShield := mgr.targets[target][0].hp
+	index := 0
+	for i, shield := range mgr.targets[target] {
+		if shield.hp > maxShield {
+			maxShield = shield.hp
+			index = i
+		}
+	}
+	return index
+}
```
unleashurgeek, line 56: * The current implementation will cause the sim to crash (and does for our integration tests). Consider the case when the target has no shields (length of `activeShields` is 0 so getting a shield at index 0 is impossible)
* This function is named with a capital letter (which makes it public). I think that makes sense, but the return does not make sense for a public function. If I am some other service and I call `MaxShield`, returning the index tells me nothing since I have no way to access shields by index (and probably shouldn't). Given how `MaxShield` is currently used, I think it makes more sense for this to just return the `hp` of the max shield (or `0` if the target has no shields)

pkg/engine/event/shield.go
```diff
@@ -21,8 +21,8 @@ type ShieldRemoved struct {
...
 type ShieldChangeEventHandler = handler.EventHandler[ShieldChange]
 type ShieldChange struct {
-	ID     key.Shield   `json:"id"`
-	Target key.TargetID `json:"target"`
-	OldHP  float64      `json:"old_hp"`
-	NewHP  float64      `json:"new_hp"`
+	Target    key.TargetID `json:"target"`
+	NewHP     float64      `json:"new_hp"`
+	DamageIn  float64      `json:"damage_in"`
+	DamageOut float64      `json:"damage_out"`
```
unleashurgeek, line 27: thinking out loud: should we also return the `key.Shield` of the max shield or do you think that is unnecessary information?

pkg/engine/shield/absorb.go
```diff
@@ -1,37 +1,63 @@
...
+		remaining := math.Dim(damage, shield.hp)
+		shield.hp = math.Dim(shield.hp, damage)
+		// Outgoing damage is always the lowest damage remaining after initial damage is subtracted from shields
+		if remaining < damageOut {
+			damageOut = remaining
+		}
+		// Max shield is always the highest shield hp remaining after damage is applied
+		if shield.hp > newMaxShieldHP {
+			newMaxShieldHP = shield.hp
+			maxShieldID = mgr.targets[target][j].name
```
unleashurgeek, line 33: simplify to just `maxShieldID = shield.name`, no need to reference by index

pkg/engine/shield/add.go
```diff
@@ -3,21 +3,67 @@ package shield
...
 	})
 }
+
+// Check list of shields for a matching shield
+func (mgr *Manager) CheckMatching(id key.Shield, shield info.Shield) (bool, int) {
+	for i, shields := range mgr.targets[shield.Target] {
+		if shields.name == id {
+			return true, i
+		}
+	}
```
unleashurgeek, line 67: nit: `shields -> s` or something not plural

pkg/engine/shield/add.go
```diff
@@ -3,21 +3,67 @@ package shield
...
-	// 4. add shield to mgr.targets[shield.target]
-	// 5. emit ShieldAdded event
-
-	// emit to signify shield added
+	// Get the stats for the source and target
+	source := mgr.attr.Stats(shield.Source)
+	maxShield := 0.0
+	if mgr.IsShielded(shield.Target) {
+		maxShield = mgr.MaxShield(shield.Target)
+	}
```
unleashurgeek, line 17: no need to check if shielded, `MaxShield` should return 0 if the target is unshielded

unleashurgeek (approved):

PR #156, opened 2023-07-15
Testframe: abstracted test-instance Characters

iCarus0712:


tests/teststub/char_checkers.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	diff := expected / actual
+	return 0.99 < diff && diff < 1.01
+}
+
+func (c *Character) Equal(statKey prop.Property, expected float64) {
+	if require(expected, c.Stats().GetProperty(statKey)) {
+		return
+	}
+	c.t.Fatalf("Assert Fail, stat %s, expected %f, actual %f", statKey.String(), expected, c.Stats().GetProperty(statKey))
```
unleashurgeek, line 41: Should we just replace all these assert cases with testify assert calls which will also format error messages for us. IE has stuff like `InEpsilon` and `InDelta` (the asserts also do not force failures now by default so can have multiple fail in a single test):
* https://pkg.go.dev/github.com/stretchr/testify/assert#InEpsilon
* https://pkg.go.dev/github.com/stretchr/testify/assert#InDelta

tests/teststub/characters.go
```diff
@@ -75,3 +71,16 @@ func (s *Characters) getCharacterEval(idx int) testeval.ActionEval {
...
+	if len(s.testChars) < len(characters) { // eval loaded config, need to populate testchars
+		for i := range s.cfg.Characters {
+			s.testChars = append(s.testChars, newCharacter(s.cfg.Characters[i]))
+		}
+	}
+	for i := range s.testChars {
+		s.testChars[i].key = characters[i]
+		s.testChars[i].stat = s.attributes.Stats
+		s.testChars[i].t = s.t
+	}
```
unleashurgeek, line 85: do we need to reset the `model` here (not looking at how this is being used)

unleashurgeek (approved):

tests/teststub/characters.go
iCarus0712: not necessary as only the Key is relevant

PR #165, opened 2023-07-18
night on the milky way

rayeva86:
LC implemented but it actually compiles lol


internal/lightcone/erudition/nightonthemilkyway/nightonthemilkyway.go
```diff
@@ -0,0 +1,81 @@
...
+		atkBuff(engine, owner, atkAmt)
+	})
+	engine.Events().StanceBreak.Subscribe(func(e event.StanceBreak) {
+		breakDmgBonus(engine, owner, dmgAmt)
+	})
+	engine.AddModifier(owner, info.Modifier{
+		Name:   NightontheMilkyWay,
+		Source: owner,
+		State:  Amts{atkB: atkAmt, dmg: dmgAmt},
+	})
```
unleashurgeek, line 54: This `AddModifier` call does not seem to do anything (this modifier doesnt need to be registered on create because it provides no `Stats` and does not have any listeners mapped to it)

PR #166, opened 2023-07-18
Implement Hidden Shadow (LC)

RyanOld:


internal/lightcone/nihility/hiddenshadow/hiddenshadow.go
```diff
@@ -0,0 +1,86 @@
...
+	mod.Engine().Attack(info.Attack{
+		Key:        shadow,
+		Targets:    target,
+		Source:     mod.Source(),
+		AttackType: model.AttackType_PURSUED,
+		DamageType: holderInfo.Element,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: state.dmgMult,
+		},
+		UseSnapshot: true,
```
unleashurgeek, line 82: Is not a snapshot hit

PR #171, opened 2023-07-19
change return to nil on err

srliao:
```go
// execute everything on the queue. After queue execution is complete, return the next stateFn
// as the next state to run. This logic will run the exitCheck on each execution. If an exit
// condition is met, will return that state instead
func (sim *Simulation) executeQueue(phase info.BattlePhase, next stateFn) (stateFn, error) {
	// always ult check when calling executeQueue
	if err := sim.ultCheck(); err != nil {
		return next, err
	}
```

Looking at this code block, I think the intention here is `return nil, err`, since err is not nil, next should be ignored by the core state loop. Was a little confused on why the return is `next` here


unleashurgeek:
Should also change the other `ultCheck` that happens inside the loop

srliao:
> Should also change the other `ultCheck` that happens inside the loop

ok fixed

unleashurgeek (approved):

PR #186, opened 2023-07-26
Implement Seele

RyanOld:


internal/character/seele/talent.go
```diff
@@ -0,0 +1,73 @@
...
+			prop.QuantumPEN:       penAmt,
+		},
+	})
+	// if current turn is resurgence, don't add extra turn. otherwise add extra turn.
+	if state.c.resurgence {
+		state.c.resurgence = false
+	} else {
+		state.c.resurgence = true
+		state.c.engine.InsertAction(state.c.id)
+	}
```
unleashurgeek, line 72: Insertion should only happen in the following:
1. Insertion only ever occurs `OnAfterAction`
2. Can only insert iff the action was from Seele (attack, skill, or ult) and triggered a kill
3. If damage was not from any of these cases (such as pursued E6 damage), do not trigger
4. Once a Seele insert starts, can no longer insert new actions until the current turn ends/next turn starts

RyanOld: i've been trying to figure this out for a while but about no. 2 though, how do i check for kill trigger inside onAfterAction listener/actionEnd subscriber? can i nest subscriber inside a subscriber like this ? :
```go
c.engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
    c.engine.Events().TargetDeath.Subscribe(func(event event.TargetDeath) {
        // extra turn logics
    } 
}
```

i thought about like using a proc var and set it to true `onAfterAction` and check for it on `TargetDeath` but i don't exactly know the order the events

internal/character/seele/talent.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (c *char) initTalent() {
+	// listen to death events. If killer is seele,
+	// enter buffed state and set trigger for extra/resurgence turn.
+	c.engine.Events().TargetDeath.Subscribe(func(e event.TargetDeath) {
+		if e.Killer == c.id {
+			c.enterBuffedState()
+		}
+		c.hasKilled = true
```
unleashurgeek, line 36: * Should only set to true if seele is the killer
* This has a problem where if the E6 kills will trigger the buffed state regardless of who's action it is. Not sure what the best solution is at the moment. One idea is you could also have an `ActionStart` subscription that keeps track of the action owner in char and only add buffed state iff `actionStart.Owner == e.Killer == c.id`. Can also add a check of `!e.Insert` here which makes it more game accurate and simplify `ActionEnd` subscription to only look at `hasKilled` boolean (which may want to rename to something more like `shouldInsert`), but imo not strictly necessary.

For 2nd item: I am fine with not fixing it right now and cutting an issue for us to fix it in the future (but obviously best if it just works correctly)

internal/character/seele/talent.go
```diff
@@ -0,0 +1,66 @@
...
+	})
+
+	c.engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
+		// insert resurgence turn only if action is seele's, hasKilled is triggered,
+		// and not already in an insert/resurgence turn.
+		if e.Owner == c.id && c.hasKilled {
+			c.hasKilled = false
```
unleashurgeek, line 43: should always reset `hasKilled` on `ActionEnd` so kills from a prior action trigger in the next. For the case when `hasKilled` is true but `e.Owner != c.id`, it does not currently reset `hasKilled`

internal/character/seele/talent.go
```diff
@@ -0,0 +1,70 @@
...
+			c.enterBuffedState()
+		}
+	})
+
+	// actionStart subs to enable adding buffedState mod effectively after-kill
+	c.engine.Events().ActionStart.Subscribe(func(e event.ActionStart) {
+		if !e.IsInsert {
+			c.actionOwner = e.Owner
+		}
+	})
```
unleashurgeek, line 43: You need to keep track of if this action is an insert (or at least have some way to check in the `TargetDeath` subscription). Consider the following scenario:
1. Seele ults (ults are not inserts so `actionOwner = seele`)
2. The ult kills, `TargetDeath` causes seele to enter buff state and `shouldInsert` go true
3. At the end of the ult (`ActionEnd`), then `InsertAction` happens
4. (assuming queue only contains the inserted action), `ActionStart` happens for the insert. Because action is an insert, actionOwner is unchanged (`actionOwner = seele` from ult)
5. The insert kills, `TargetDeath` causes seele to enter buff state and `shouldInsert` go true
6. At the end of the ult (`ActionEnd`), then `InsertAction` happens
7. above repeats until Seele doesnt get a kill

internal/character/seele/talent.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// extra turn logic
+	c.engine.Events().ActionEnd.Subscribe(func(e event.ActionEnd) {
+		if c.shouldInsert {
+			c.shouldInsert = false
+			c.engine.InsertAction(c.id)
+		}
+		c.shouldInsert = false
+	})
```
unleashurgeek, line 52: nit: `shouldInsert` being set false twice is redundant. Delete case inside if statement.

internal/character/seele/trace.go
```diff
@@ -53,6 +53,8 @@ func (c *char) advanceForward(key key.Reason, isInsert bool) {
 	// if not insert : modifygaugecost -20%
 	// if insert : modifynormalized -20%
 	if isInsert {
+		// TODO : create issue for adding TurnOwner helper func to engine
+		// TODO : if current turn is seele's, delay modifyGaugeNormalized to OnTurnEnd.
```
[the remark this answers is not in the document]
RyanOld: Add TurnOwner to engine : issue raised at #210

unleashurgeek (approved):

PR #189, opened 2023-07-26
Implemented Blade

Kisa75:
Some TODOs remain


internal/character/blade/ult.go
```diff
@@ -0,0 +1,64 @@
...
+			model.DamageFormula_BY_MAX_HP: ultSingleHP[c.info.UltLevelIndex()],
+		},
+		DamageValue:  hpTally * (ultSingleTally[c.info.UltLevelIndex()] + e1TallyMod),
+		StanceDamage: 60.0,
+		EnergyGain:   5.0,
+	})
+
+	// Adjacent Targets
+	c.engine.Attack(info.Attack{
+		Key:        Ult,
```
unleashurgeek, line 50: nit: probably want separate key for primary and adjacent (I created separate keys for arlan ult since different scalings)

internal/character/blade/technique.go
```diff
@@ -0,0 +1,35 @@
+package blade
+
+import (
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	Technique key.Attack = "blade-technique"
```
unleashurgeek, line 10: untype key to remove need to cast

internal/character/blade/talent.go
```diff
@@ -0,0 +1,66 @@
...
+				DamageType: model.DamageType_WIND,
+				AttackType: model.AttackType_INSERT,
+				BaseDamage: info.DamageMap{
+					model.DamageFormula_BY_ATK:    talentAtk[c.info.TalentLevelIndex()],
+					model.DamageFormula_BY_MAX_HP: (talentHP[c.info.TalentLevelIndex()] + e6AddMod),
+				},
+				StanceDamage: 30.0,
+				EnergyGain:   10.0,
+			})
+		},
```
unleashurgeek, line 40: * 3 hit attack, 33% hit split ratio
* `AttackEnd` after the three hits
* Heal happens after the attack
* Charge count resets after the heal (its possible for blade to hit an edge case where the insert gets cancelled before it happens. In this instance charge count does not get reset).

internal/character/blade/ult.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+
+	hpTally := math.Min(c.hpLoss, 0.9*c.engine.Stats(c.id).MaxHP())
+
+	// TODO: Seperate Tally?
+	e1TallyMod := 0.0
+
+	if c.info.Eidolon >= 1 {
+		e1TallyMod = 1.5
+	}
```
unleashurgeek, line 30: this looks good to me

internal/character/blade/attack.go
```diff
@@ -0,0 +1,83 @@
...
+				},
+				StanceDamage: 60.0,
+				EnergyGain:   30.0,
+				HitRatio:     hitRatio,
+			})
+		}
+
+		// Adjacent Targets
+		c.engine.Attack(info.Attack{
+			Key:        EnhancedNormal,
```
unleashurgeek, line 71: May want to separate into two separate keys

internal/character/blade/blade.go
```diff
@@ -0,0 +1,122 @@
...
+	})
+}
+
+type char struct {
+	engine    engine.Engine
+	id        key.TargetID
+	info      info.Character
+	hpLoss    float64
+	charge    int
+	maxCharge int
```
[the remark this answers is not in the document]
Kisa75: Agreed to keep the existing implementation

internal/character/blade/blade.go
```diff
@@ -0,0 +1,119 @@
...
+		if !c.engine.HasModifier(c.id, GainedCharge) {
+			c.engine.AddModifier(c.id, info.Modifier{
+				Name:     Talent,
+				Source:   c.id,
+				MaxCount: float64(c.maxCharge),
+			})
+			c.engine.AddModifier(c.id, info.Modifier{
+				Name:   GainedCharge,
+				Source: c.id,
+			})
```
unleashurgeek, line 96: shouldn't `charge++` also happen here instead of adding `Talent` modifier?

internal/character/blade/blade.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	} else {
+		c.charge++
+	}
+
+	if c.charge >= c.maxCharge {
+		c.Talent()
+	}
```
unleashurgeek, line 104: I think should only make this call when `c.charge++` is called. In the whole charge increase logic above, might want to limit it to only run when not yet at max charges so `Talent()` only gets called once

unleashurgeek (approved):

PR #190, opened 2023-07-27
Implement Memories of the Past (LC)

RyanOld:


internal/lightcone/harmony/memoriesofthepast/memoriesofthepast.go
```diff
@@ -0,0 +1,80 @@
...
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_HARMONY,
+		Promotions:    promotions,
+	})
+	modifier.Register(memories, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAfterAttack: addEnergy,
+			// DM uses OnListenTurnEnd. try to imitate with normal and skill action check.
+			OnAfterAction: resetCooldown,
```
unleashurgeek, line 37: `OnAfterAction` is not equivalent to `OnListenTurnEnd`. Anything that starts with `OnListen` in the datamine means it triggers on **any** event rather than events associated with the character (which is what `OnAfterAction` would be). Since Memories of the Past uses `OnListenTurnEnd`, this means that the energy gain will reset on **any** turn and not just the turn of the equipped character (so can get energy from any ults or insert attacks that occur outside of the character's turn)

The best way to do this would be with event subscriptions (event subscriptions are our equivalent of `OnListen*`). I would say implement similar to "Time Waits for No One" which uses subscriptions for the logic. That is:
* `TurnEnd` subscription which resets the cooldown
* `AttackEnd` subscription that filters for only attacks performed by the LC owner, which on those attacks (if not on CD) give energy and give CD.
* 
You could also have the CD stored as a modifier for better traceability in the sim

internal/lightcone/harmony/memoriesofthepast/memoriesofthepast.go
```diff
@@ -0,0 +1,69 @@
...
+	lightcone.Register(key.MemoriesofthePast, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_HARMONY,
+		Promotions:    promotions,
+	})
+	modifier.Register(memories, modifier.Config{})
+
+	modifier.Register(memoriesCD, modifier.Config{
+		Stacking: modifier.ReplaceBySource,
```
unleashurgeek, line 32: nit: this stacking behavior is not necessary

internal/lightcone/harmony/memoriesofthepast/memoriesofthepast.go
```diff
@@ -0,0 +1,69 @@
...
+	})
+
+	// Resets CD on all turn end
+	engine.Events().TurnEnd.Subscribe(func(e event.TurnEnd) {
+		engine.RemoveModifier(owner, memoriesCD)
+	})
+
+	// add energy on all attack end by lc holder (for ults etc.)
+	engine.Events().AttackEnd.Subscribe(func(event event.AttackEnd) {
+		if engine.HasModifier(owner, memoriesCD) || event.Attacker != owner {
```
unleashurgeek, line 53: flip the order here so owner check happens before cd check (more optimal this way)

internal/lightcone/harmony/memoriesofthepast/memoriesofthepast.go
```diff
@@ -0,0 +1,69 @@
...
+			Key:    memories,
+			Target: owner,
+			Source: owner,
+			Amount: energyAmt,
+		})
+		// enter cd
+		engine.AddModifier(owner, info.Modifier{
+			Name:     memoriesCD,
+			Source:   owner,
+			Duration: 1,
```
unleashurgeek, line 66: Remove the duration (has to be explicitly removed to turn off)

unleashurgeek (approved):

PR #193, opened 2023-08-06
March7th

aaron-hwang:
A little stuck on talent implementation details; specifically handling the count comparisons (Checking whether March can still counter attack). E4 also not fully implemented as a result.


aaron-hwang:
left comments regarding talent in the file

internal/character/march7th/eidolon.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+		},
+	})
+
+	modifier.Register(MarchE2, modifier.Config{
+		Stacking: modifier.Replace,
+		Listeners: modifier.Listeners{
+			OnAdd: determineE2Target,
+		},
+	})
```
unleashurgeek, line 40: This does not do what you want. `OnAdd` just runs when the modifier is added. How you are using it will just run immediately when `AddModifier` is called.

Instead of using a modifier for this, should just use a `BattleStart` subscription to defer the execution until when the battel starts (since need to wait for all characters to be created first)

internal/character/march7th/eidolon.go
```diff
@@ -0,0 +1,82 @@
...
+			lowestHpRatio = mod.Engine().HPRatio(Target)
+		}
+	}
+	e2Target := mod.Engine().Retarget(info.Retarget{
+		Targets: mod.Engine().Characters(),
+		Max:     1,
+		Filter: func(target key.TargetID) bool {
+			return mod.Engine().HPRatio(target) == lowestHpRatio
+		},
+	})[0]
```
unleashurgeek, line 76: If you're going to loop over all characters to find the lowest hp ratio, why even call retarget and just get/save the target id from the first loop?

internal/character/march7th/skill.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// A4 check
+	shieldDur := 3
+	if c.info.Traces["102"] {
+		shieldDur += 1
+	}
+
+	c.engine.AddModifier(target, info.Modifier{
+		Name:   Skill,
```
unleashurgeek, line 78: nit: I'd rename this to `march7th-shield`

internal/character/march7th/skill.go
```diff
@@ -0,0 +1,93 @@
...
+				if mod.Engine().Stats(mod.Owner()).CurrentHPRatio() >= 30 {
+					mod.AddProperty(prop.AggroPercent, 5)
+				}
+			},
+			OnRemove: func(mod *modifier.Instance) {
+				mod.Engine().RemoveShield(Skill, mod.Owner())
+			},
+			OnPhase1: func(mod *modifier.Instance) {
+				march7th, _ := mod.Engine().CharacterInfo(mod.Source())
+				if march7th.Eidolon >= 6 {
```
unleashurgeek, line 44: rather than calling `CharacterInfo`, I'd pass the heal stats as part of the state. If E6 is not enabled, just set the heal value and heal percent to 0 in the state

internal/character/march7th/skill.go
```diff
@@ -0,0 +1,93 @@
...
+	E6heal = "march7th-e6hot"
+)
+
+type shieldState struct {
+	shieldPercentage float64
+	shieldFlat       float64
+}
+
+func init() {
+	modifier.Register(Skill, modifier.Config{
```
unleashurgeek, line 22: Status Type `BUFF`

internal/character/march7th/talent.go
```diff
@@ -0,0 +1,141 @@
...
+	modifier.Register(Talent, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAdd: func(mod *modifier.Instance) {
+				for _, teammate := range mod.Engine().Characters() {
+					mod.Engine().AddModifier(teammate, info.Modifier{
+						Name:   MarchAllyMark,
+						Source: mod.Owner(),
+					})
+				}
+			},
```
unleashurgeek, line 61: Same comment with E2, not how `OnAdd` works

internal/character/march7th/talent.go
```diff
@@ -0,0 +1,141 @@
...
+When March joins batttle, she is given a modifier that represents her actual talent.
+This modifier, when added, adds another modifier to all of March's teammates, which I called MarchAllyMark
+MarchAllyMark listens for whenever the person it is attached to is about to be attacked. If at least one of the targets in
+that attack is shielded, AND the attacker is an enenmy, add a modifier called MarchCounterMark to the attacker
+MarchCounterMark listens for when whoever it is attached to is done attacking.
+Afterwards, it tells March to counter attack. If successful, March will counter, and (?) remove
+one stack from the modifier that determines if march can counter or not
+MarchCounterMark will not be added to an enemy if March does not have enough counters left.
+March regenerates her counters on Phase2End.
+*/
```
unleashurgeek, line 31: Instead of using modifiers, majority of this logic is a better fit for subscriptions. A subscription runs on every event of that type rather than attached to a specific target. For cases like the "MarchAllyMark", this allows you to get every `AttackStart` event and then filter down to only the ones you care about (unlike the game, we can write code and do not have to be limited to modifiers).

Thinking out loud on how to implement this talent:
* `AttackStart` subscription that does the logic of `MAvatar_March7th_00_BPSkill_Shield_Counter` (will need to loop through list of targets)
* Can keep counter as either a modifier or variable in the char struct. A variable is easier to implement, a modifier is better for data traceability (char variables are not logged so harder for sim users to trace)
* Add mark modifier to enemies from the `AttackStart` subscription which will insert the counter `OnAfterAttack` (equivalent of `MAvatar_March7th_00_BPSkill_Shield_Mark`)
* Can still have talent modifier with `OnPhase2` subscription to reset stacks (need to also make sure that character starts with stacks)

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,99 @@
...
+					Amount: 6,
+				})
+			}
+		}
+	}
+}
+
+func (c *char) ultHit(hitIndex int, hitRatio float64) {
+	hitTargets := make(map[key.TargetID]bool)
+	for i := 0; i < 2; i++ {
```
unleashurgeek, line 63: should loop 3 times, currently only loop twice

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,99 @@
...
+					Source: c.id,
+					Amount: 6,
+				})
+			}
+		}
+	}
+}
+
+func (c *char) ultHit(hitIndex int, hitRatio float64) {
+	hitTargets := make(map[key.TargetID]bool)
```
unleashurgeek, line 62: nit: can do the following to preallocate the map which is more performant
```go
hitTargets := make(map[key.TargetID]bool, len(c.engine.Enemies()))
```

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,99 @@
...
+			HitIndex:     hitIndex,
+			HitRatio:     hitRatio / 3,
+			StanceDamage: 60,
+			EnergyGain:   0,
+			DamageType:   model.DamageType_ICE,
+			AttackType:   model.AttackType_ULT,
+			BaseDamage: info.DamageMap{
+				model.DamageFormula_BY_ATK: ult[c.info.UltLevelIndex()],
+			},
+		})
```
unleashurgeek, line 86: hit ratio should always be `25%` since `ultHit` will only ever hit a target at most once

internal/character/march7th/technique.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	freezeTarget := c.engine.Retarget(info.Retarget{
+		Targets: c.engine.Enemies(),
+		Filter: func(target key.TargetID) bool {
+			return c.engine.Stats(target).GetDebuffRES(model.BehaviorFlag_STAT_CTRL_FROZEN) <= 0.8 &&
+				c.engine.Stats(target).GetDebuffRES(model.BehaviorFlag_STAT_CTRL) <= 0.8
+		},
```
unleashurgeek, line 16: `GetDebuffRES` takes in multiple flags by design (varags param) for cases such as this where there are multiple being searched. In this case, only the highest value is taken. We want to avoid calling `Stats` multiple time since it is an expensive operation.

```go
Filter: func(target key.TargetID) bool {
	return c.engine.Stats(target).GetDebuffRES(
      model.BehaviorFlag_STAT_CTRL_FROZEN, model.BehaviorFlag_STAT_CTRL) <= 0.8
},
```

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,99 @@
...
+		if c.info.Eidolon >= 1 {
+			if successfullyApplied {
+				c.engine.ModifyEnergy(info.ModifyAttribute{
+					Key:    Ult,
+					Target: c.id,
+					Source: c.id,
+					Amount: 6,
+				})
+			}
+		}
```
unleashurgeek, line 57: Want to move `ModifyEnergy` out of for loop to only be called once like datamine in the event that some other mechanic subscribes to energy change events and is based on event count.

internal/character/march7th/eidolon.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (c *char) addE2Shield(e event.BattleStart) {
+	lowestHpRatio := 1.0
+	var lowestHpTarget key.TargetID
+	for _, Target := range c.engine.Characters() {
+		if c.engine.HPRatio(Target) <= lowestHpRatio {
+			lowestHpRatio = c.engine.HPRatio(Target)
+			lowestHpTarget = Target
+		}
+	}
```
unleashurgeek, line 45: since not using retarget, need to manually exclude targets in limbo (targets that have 0 HP).

internal/character/march7th/march7th.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewInstance(engine engine.Engine, id key.TargetID, charInfo info.Character) info.CharInstance {
+	c := &char{
+		engine: engine,
+		id:     id,
+		info:   charInfo,
+	}
+	c.initTalent()
+
+	engine.Events().BattleStart.Subscribe(c.addE2Shield)
```
unleashurgeek, line 54: missing E2 check. Current implementation will always add E2 shield regardless if E2 is enabled or not

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,102 @@
...
+const (
+	MarchUltFreeze = "march7th-ultfreeze"
+	Ult            = "march7th-ult"
+)
+
+func init() {
+	modifier.Register(MarchUltFreeze, modifier.Config{
+		Duration: 1,
+	})
+}
```
unleashurgeek, line 20: Remove `MarchUltFreeze` since it is unused

internal/character/march7th/ult.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, t := range targets {
+			hitTargets[t] = true
+		}
+	}
+
+	c.engine.ModifyEnergy(info.ModifyAttribute{
+		Key:    Ult,
+		Target: c.id,
+		Source: c.id,
+		Amount: 5 * 0.25,
```
unleashurgeek, line 100: nit: for consistency, should be `5 * hitRatio`

internal/character/march7th/skill.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	Skill  = "march7th-shield"
+	E6heal = "march7th-e6hot"
```
unleashurgeek, line 13: nit: `E6heal = "march7th-e6hot"` -> `E6 = "march7th-e6"`

internal/character/march7th/skill.go
```diff
@@ -0,0 +1,103 @@
...
+					Source: mod.Source(),
+					Target: mod.Owner(),
+					BaseShield: info.ShieldMap{
+						model.ShieldFormula_SHIELD_BY_SHIELDER_DEF: mod.State().(shieldState).shieldPercentage,
+					},
+					ShieldValue: mod.State().(shieldState).shieldFlat,
+				})
+				if mod.Engine().Stats(mod.Owner()).CurrentHPRatio() >= 30 {
+					mod.AddProperty(prop.AggroPercent, 5)
+				}
```
unleashurgeek, line 40: `HPRatio` is a value between `[0, 1]`. Should be `>= 0.30`

internal/character/march7th/talent.go
```diff
@@ -0,0 +1,123 @@
...
+	})
+}
+
+func (c *char) applyCounterMark(e event.AttackStart) {
+	hasShieldedTarget := false
+	for _, ally := range c.engine.Characters() {
+		hasShieldedTarget = hasShieldedTarget || c.engine.IsShielded(ally)
+	}
+
+	haveCounters := c.engine.HasModifier(c.id, TalentCount)
```
unleashurgeek, line 78: Instead of calling `HasModifier` use `ModifierStackCount` and check if count > 0. There are potential edge cases where the stack count will be 0 but modifier will still be active

internal/character/march7th/talent.go
```diff
@@ -0,0 +1,123 @@
...
+				DamageType: model.DamageType_ICE,
+				BaseDamage: info.DamageMap{
+					model.DamageFormula_BY_ATK: talent[march7th.TalentLevelIndex()],
+					model.DamageFormula_BY_DEF: e4Scaling,
+				},
+			})
+			// Remove a count/stack from the talent counter
+			mod.Engine().ExtendModifierCount(mod.Source(), TalentCount, -1)
+			// Remove this modifier from the enemy it is attached to
+			mod.Engine().RemoveModifier(mod.Owner(), MarchCounterMark)
```
unleashurgeek, line 116: Remove this modifier after the `InsertAbility` call rather than inside the `Execute` to match game logic

unleashurgeek (approved):

PR #198, opened 2023-08-14
Implement Planetary Rendezvous (LC)

RyanOld:


internal/lightcone/harmony/planetaryrendezvous/planetaryrendezvous.go
```diff
@@ -0,0 +1,59 @@
...
+			if charInfo.Element == holderInfo.Element {
+				// if element matches. add perm element specific dmg buff.
+				engine.AddModifier(char, info.Modifier{
+					Name:   dmgBuff,
+					Source: owner,
+					Stats:  statsMap[holderInfo.Element],
+				})
+			}
+		}
+	})
```
unleashurgeek, line 58: This current behavior is wrong. This current logic is only buffing characters that share the same element as the owner. Instead all allies should receive the buff of element type (think of Tingyun where she can cause any ally to do electric damage regardless of their type).

Recommended changes:
1. Instead of using the `statsMap`, use the `prop.DamagePercent` function to map `DamageType` to `DamagePercent`
2. simplify the `BattleStart` logic to just apply the buff to all characters

```go
dmgAmt := 0.09 + 0.03*float64(lc.Imposition)

engine.Events().BattleStart.Subscribe(func(event event.BattleStart) {
  holderInfo, _ := engine.CharacterInfo(owner)
  for _, char := range engine.Characters() {
    engine.AddModifier(char, info.Modifier{
      Name: dmgBuff,
      Source: owner,
      Stats: {prop.DamagePercent(holderInfo.Element): dmgAmt},
    })
  }
}
```

pkg/key/lightcone.go
```diff
@@ -61,6 +61,7 @@ const (
 	CarvetheMoonWeavetheClouds LightCone = "carve_the_moon_weave_the_clouds"
 	MemoriesofthePast          LightCone = "memories_of_the_past"
 	DanceDanceDance            LightCone = "dance_dance_dance"
+	PlanetaryRendezvous        LightCone = "planetary-rendezvous"
```
unleashurgeek, line 64: underscore instead of dash

internal/lightcone/harmony/planetaryrendezvous/planetaryrendezvous.go
```diff
@@ -0,0 +1,45 @@
...
+// DMG dealt increases by 12%.
+
+func init() {
+	lightcone.Register(key.PlanetaryRendezvous, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        4,
+		Path:          model.Path_HARMONY,
+		Promotions:    promotions,
+	})
+	modifier.Register(dmgBuff, modifier.Config{})
```
unleashurgeek, line 28: The stacking behavior of this should be `ReplaceBySource` to allow multiple being equipped at once.

unleashurgeek (approved):

Issue #187, opened 2023-07-26
Add `TurnOrder` method to engine


unleashurgeek:
Needed for Past and Future and Bronya LC. Should be a function that returns an `[]key.TargetID` which is a copy of the current turn order. Implementations can then filter this down to find their desired target based on the current turn order.

PR #200, opened 2023-08-17
complete Turn Manager, add TurnOrder

agentyoda:
#187 implemented


pkg/engine/turn/turn.go
```diff
@@ -123,67 +140,76 @@ func (mgr *manager) RemoveTarget(id key.TargetID) {
...
 	mgr.gaugeCost = 1.0
 	mgr.activeTurn = true
 	mgr.activeTarget = mgr.order.order[0].id
 	av := mgr.av(mgr.activeTarget)
 
-	// mgr.order.order[0].gauge = 0 // set gauge/av of active to 0
-
 	for _, t := range mgr.order.order {
 		t.gauge -= av * mgr.attr.Stats(t.id).SPD()
 	}
```
unleashurgeek, line 165: Thinking out loud: I wonder if we need to do a stable sort prior to the logic here. AV is determined by two factors
1. gauge, which the turn manager is responsible for and will re-sort any time it changes
2. speed, which turn manager is **not** responsible for and may change between sorts. In the event that speed changes between the last sort and start of turn, the target at the top of the order may not be the one with the lowest AV.

pkg/engine/turn/turn.go
```diff
@@ -123,67 +140,76 @@ func (mgr *manager) RemoveTarget(id key.TargetID) {
...
 			"target at top of order must have 0 gauge to call reset (their turn is active) %+v", mgr.order.order[0])
 	}
+
 	mgr.activeTurn = false
 	mgr.order.order[0].gauge = BaseGauge * mgr.gaugeCost
+	
+	// It would be more efficient to loop through mgr.order ourselves to determine this single target's placement instead of resorting the whole array when no other elements are changing.
+	// Unless we are also checking for other SPD changes that happened during the turn, in which case sort.Stable() is better to use, but only after we move the element to the end
+	// so as to ensure that, in the case of a tie, it is properly at the tail end of the tied elements.
+	mgr.order.order = append(mgr.order.order[1:], mgr.order.order[0])
```
unleashurgeek, line 193: nit: might want to rename the variables since `mgr.order.order` is not clear what it does

pkg/engine/turn/turn.go
agentyoda: The only thing that appears to change between the last check (in ResetTurn, called in phase2) and beginTurn calling StartTurn are phase2's ActionEnd and ModifierPhase2 modifier ticks. Would those include any SPD-changing effects? I had assumed those would not, since active SPD-buffing skills are called in 'action' to my knowledge, but I suppose that phase2 might include things like SPD buffs expiring?

Even if that's not the case, just to future-proof it, so that even if there aren't any SPD-changing effects now but might be later, I'll go ahead and add a sort.

agentyoda: Though I'm wondering: would any of those phase2 effects depend on our gauge reset done in ResetTurn? If not, we could also just call ResetTurn last in phase2, as then nothing would change between phase2 and beginTurn, to my knowledge, so we could be a tiny bit more efficient and not call an extra Sort.

pkg/engine/turn/turn.go
agentyoda: Renamed.

pkg/engine/turn/turn.go
unleashurgeek: >The only thing that appears to change between the last check (in ResetTurn, called in phase2) and beginTurn calling StartTurn are phase2's ActionEnd and ModifierPhase2 modifier ticks

there is also the `executeQueue` call, which will run the logic for any queued up actions and inserts (counters, follow up attacks, ults, extra turns/actions). This ends up being a lot of things. An example of a SPD modifier that happens here would be Asta ult if queued after the active turn action is performed.

>Though I'm wondering: would any of those phase2 effects depend on our gauge reset done in ResetTurn?

Sadly yes, reset needs to happen after the action and prior to/at the start of phase2. Game has odd gauge mechanics based on if you run before or after the reset. A good example is the advance on Sushang ult. It works by setting her gauge to 0 when the ult is performed. In game, if the ult is performed at the start of her turn the advance does nothing/does not happen because her gauge is currently 0. When the reset happens, it will be 10k. If you ult after her action but still on her turn in game, that will cause the advance to work and will result in the next turn also being Sushang. This is because the reset happens immediately after the turn action is performed rather than at the end of the turn

agentyoda: Ok, that's interesting; I assumed queued ults would just be coded as new turns. In light of the reset stuff, I'll go ahead and add another sort to StartTurn.

pkg/engine/turn/modify.go
```diff
@@ -1,48 +1,64 @@
...
-	//		turn (so index 1 instead of 0 when 0 gauge/AV)
-	// 4. emit GaugeChangeEvent
+	// find target index in mgr.orderHandler.turnOrder
+	targetIndex := 0
+	for i, v := range mgr.orderHandler.turnOrder {
+		if data.Target == v.id {
+			targetIndex = i
+			break
+		}
+	}
```
unleashurgeek, line 34: move this to a function to cleanup usage. Same index discover exists in `RemoveTarget`. May want to update `Target(key.TargetID)` to return both the index and `*Target` since it is also a similar loop?

Should also account for case where target is not found and throw an error (here and `RemoveTarget`)

pkg/engine/turn/modify.go
```diff
@@ -1,48 +1,64 @@
...
+		} else {
+			mgr.orderHandler.turnOrder = append([]*target{mgr.target(data.Target)}, mgr.orderHandler.turnOrder[:targetIndex]...)
+		}
+	} else {
+		if targetIndex + 1 < mgr.orderHandler.Len() {
+			mgr.orderHandler.turnOrder = append([]*target{mgr.orderHandler.turnOrder[0]}, append([]*target{mgr.target(data.Target)}, append(mgr.orderHandler.turnOrder[1:targetIndex], mgr.orderHandler.turnOrder[targetIndex+1:]...)...)...)
+		} else {
+			mgr.orderHandler.turnOrder = append([]*target{mgr.orderHandler.turnOrder[0]}, append([]*target{mgr.target(data.Target)}, mgr.orderHandler.turnOrder[1:targetIndex]...)...)
 		}
 	}
```
unleashurgeek, line 51: My suggestion is change this to the following steps:
1. Set `turnOrder` to be a new array with the target removed (`append(mgr.orderHandler.turnOrder[:targetIndex], mgr.orderHandler.turnOrder[targetIndex+1:]...)`)
2. Add the target to the top of the array
3. flip index 0 and index 1 iff `mgr.activeTarget != data.Target`

Alternatively, can go with a solution that only needs to loop over the array once and shuffles everything down in-place. Would look something like this (not tested, may not be correct):
```go
startIndex := 0
if mgr.activeTurn && targetIndex != 0 {
  startIndex = 1
}

prev := mgr.orderHandler.turnOrder[targetIndex] // use mgr.target return if saved from another call
for i := startIndex; i <= targetIndex; i++ {
  tmp := mgr.orderHandler.turnOrder[i]
  mgr.orderHandler.turnOrder[i] = prev
  prev = tmp
}
```

pkg/engine/turn/modify.go
```diff
@@ -1,48 +1,64 @@
...
-	prev := mgr.target(data.Target).gauge
+
+	previousGauge := mgr.target(data.Target).gauge
 	mgr.target(data.Target).gauge = data.Amount
 
-	if mgr.target(data.Target).gauge == prev {
+	// if there's no change to Gauge, exit early
+	if mgr.target(data.Target).gauge == previousGauge {
 		return nil
 	}
```
unleashurgeek, line 25: best to minimize this code to only call `mgr.target(data.Target)` once since that is a for loop over all targets every time it is called

unleashurgeek (commented):
need to fix linter errors (most errors are because of unused import causing compilation to fail)

PR #203, opened 2023-08-29
Write AddShield() unit tests

GortFan:


pkg/engine/shield/add_test.go
```diff
@@ -0,0 +1,362 @@
...
+				Source:     source,
+				Target:     target,
+				BaseShield: info.ShieldMap{model.ShieldFormula_SHIELD_BY_SHIELDER_TOTAL_SHIELD: 0.5},
+			},
+		},
+	}
+
+	manager.event.ShieldAdded.Subscribe(func(event event.ShieldAdded) {
+		assert.Fail(t, "A shield of 0 hp should not be added to the target, as such this event should never emit")
+	})
```
unleashurgeek, line 252: This assert is failing (right now it is possible to add 0 HP shields and they do emit events)

pkg/engine/shield/add_test.go
```diff
@@ -0,0 +1,362 @@
...
+	shieldID := key.Shield("MockShield")
+	shieldInfo := info.Shield{
+		Source:     source,
+		Target:     target,
+		BaseShield: info.ShieldMap{model.ShieldFormula_SHIELD_BY_SHIELDER_TOTAL_SHIELD: 0.5},
+	}
+
+	manager.event.ShieldAdded.Subscribe(func(event event.ShieldAdded) {
+		assert.Fail(t, "A shield of 0 hp should not be added to the target, as such this event should never emit")
+	})
```
unleashurgeek, line 286: same comment about test failing

pkg/engine/shield/add_test.go
```diff
@@ -0,0 +1,362 @@
...
+				Source:     source,
+				Target:     target,
+				BaseShield: info.ShieldMap{model.ShieldFormula_SHIELD_BY_SHIELDER_TOTAL_SHIELD: 0.5},
+			},
+		},
+	}
+
+	manager.event.ShieldAdded.Subscribe(func(event event.ShieldAdded) {
+		assert.Equal(t, 50.0, event.ShieldHealth)
+	})
```
unleashurgeek, line 90: nit: should have different values for each case to make sure the wires are not crossed (right now could have `BY_TARGET_MAX_HP` using `ATK` and this test wouldn't catch it). Ideally you want 1 unit test only testing 1 thing so when it fails it is clear what is broken

pkg/engine/shield/add_test.go
```diff
@@ -0,0 +1,362 @@
...
+)
+
+// Unit Tests for AddShield()
+func TestShieldHealthByPositiveValues(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	engine := mock.NewMockEngineWithEvents(mockCtrl)
+	attr := mock.NewMockAttribute(mockCtrl)
+	manager := New(engine.Events(), attr)
+
+	defer mockCtrl.Finish()
```
unleashurgeek, line 24: nit: `defer` is not necessary (mistake in other tests)

pkg/engine/shield/add_test.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/simimpact/srsim/pkg/model"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/simimpact/srsim/tests/mock"
+)
+
+// Unit Tests for AddShield()
+func TestShieldHealthByPositiveValues(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	engine := mock.NewMockEngineWithEvents(mockCtrl)
```
unleashurgeek, line 20: `MockEngine` is not necessary here. Can just create your own event system via `events := &event.System{}`

unleashurgeek (commented):
Great start! Should add more tests than just `add` but for all shield functionality

unleashurgeek (approved):

PR #205, opened 2023-08-31
Write unit tests for Manager functions

GortFan:


pkg/engine/shield/manager_test.go
```diff
@@ -0,0 +1,121 @@
...
+	mockCtrl := gomock.NewController(t)
+	events := &event.System{}
+	attr := mock.NewMockAttribute(mockCtrl)
+	manager := New(events, attr)
+
+	target := key.TargetID(1)
+
+	isTrue := manager.HasShield(target, key.Shield("Shield"))
+	assert.Equal(t, false, isTrue)
+}
```
unleashurgeek, line 39: nit: should have case of has shields but not one of key time (`HasShield` false for non-empty shield array)

unleashurgeek (approved):

unleashurgeek (approved):

PR #211, opened 2023-09-05
Initial weakness break effects implementation

esbraff:
Needs enemies to have flag for checking if it's elite or not to be fully working


internal/global/common/weakness_break.go
```diff
@@ -0,0 +1,200 @@
...
+			Target: enemyID,
+			Source: characterID,
+			Amount: 0.25,
+		})
+
+		engine.AddModifier(enemyID, info.Modifier{
+			Name:   BreakImprisonment,
+			Source: characterID,
+			State: BreakImprisonState{
+				DelayRatio: 0.2 * (1 + engine.Stats(characterID).BreakEffect()),
```
unleashurgeek, line 194: same comment with entanglement

internal/global/common/bleed.go
```diff
@@ -56,7 +56,7 @@ func init() {
 }
 
 func bleedPhase1(mod *modifier.Instance) {
-	state, ok := mod.State().(BleedState)
+	state, ok := mod.State().(*BleedState)
```
unleashurgeek, line 59: to my pointer comment: only entanglement needed to change because was incorrect. Fine with others changing for consistency, but then you need to check the entire codebase for other places where we are creating the non break dots and make sure they are creating a dereferenced state otherwise that code will crash when ran

internal/global/common/weakness_break.go
```diff
@@ -0,0 +1,194 @@
...
+		applyWeaknessBreakIce(engine, characterID, enemyID, characterInfo, enemyInfo, maxStanceMultiplier)
+	case model.DamageType_THUNDER:
+		applyWeaknessBreakThunder(engine, characterID, enemyID, characterInfo, enemyInfo, maxStanceMultiplier)
+	case model.DamageType_WIND:
+		applyWeaknessBreakWind(engine, characterID, enemyID, characterInfo, enemyInfo, maxStanceMultiplier)
+	case model.DamageType_QUANTUM:
+		applyWeaknessBreakQuantum(engine, characterID, enemyID, characterInfo, enemyInfo, maxStanceMultiplier)
+	case model.DamageType_IMAGINARY:
+		applyWeaknessBreakImaginary(engine, characterID, enemyID, characterInfo, enemyInfo, maxStanceMultiplier)
+	}
```
unleashurgeek, line 193: given the number of params and how they are all identical, might be cleaner to construct and pass in a struct instead:

```go
type breakInfo struct {
  engine              engine.Engine
  charID              key.TargetID
  charInfo            *info.Character
  enemyID             key.TargetID
  enemyInfo           *info.Character
  maxStanceMultiplier float64
}

func ApplyWeaknessBreakEffects(engine engine.Engine, characterID, enemyID key.TargetID) {
	characterInfo, err := engine.CharacterInfo(characterID)
	if err != nil {
		panic("incorrect target id for character")
	}

	enemyInfo, err := engine.EnemyInfo(enemyID)
	if err != nil {
		panic("incorrect target id for enemy")
	}

  break := &breakInfo{
    engine:              engine,
    charID:              characterID,
    charInfo:            &characterInfo,
    enemyID:             enemyID,
    enemyInfo:           &enemyInfo,
    maxStanceMultiplier: 0.5 + enemyInfo.MaxStance/120,
  }

  switch characterInfo.Element {
  case model.DamageType_PHYSICAL:
		applyWeaknessBreakPhysical(break)
	case model.DamageType_FIRE:
		applyWeaknessBreakFire(break)
	case model.DamageType_ICE:
		applyWeaknessBreakIce(break)
	case model.DamageType_THUNDER:
		applyWeaknessBreakThunder(break)
	case model.DamageType_WIND:
		applyWeaknessBreakWind(break)
	case model.DamageType_QUANTUM:
		applyWeaknessBreakQuantum(break)
	case model.DamageType_IMAGINARY:
		applyWeaknessBreakImaginary(break)
	}
}
```

Additionally `dealWeaknessBreakDamage` can then be simplified to the signature of
```go
func dealWeaknessBreakDamage(data *breakInfo, damageMultiplier float64)
```

could also make all these functions a function of `breakInfo` to change calls to `break.applyWeaknessBreakWind()` and `break.dealWeaknessBreakDamage(dmgMultiplier float64)`

internal/global/common/weakness_break.go
```diff
@@ -0,0 +1,206 @@
...
+		State: &BreakBleedState{
+			BaseDamageValue: bleedBaseDmg,
+		},
+		Duration: 2,
+		Chance:   1.5,
+	})
+}
+
+func applyWeaknessBreakFire(breakInfoData *breakInfo) {
+	dealWeaknessBreakDamage(breakInfoData, WeaknessBreakFire, 1)
```
unleashurgeek, line 84: Fire `damageMultiplier` is `2`

unleashurgeek (approved):

PR #213, opened 2023-09-06
Implement Champion of Streetwise Boxing (Relic)

RyanOld:


internal/relic/cavern/championboxing/championboxing.go
```diff
@@ -0,0 +1,61 @@
...
+	})
+}
+
+func addAtkBuff(mod *modifier.Instance, e event.AttackEnd) {
+	mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+		Name:   atkBuff,
+		Source: mod.Owner(),
+		Stats:  info.PropMap{prop.ATKPercent: 0.05},
+	})
+}
```
unleashurgeek, line 61: nit: good to manually check count before calling add just to reduce log noise (current state will log mod added every time an attack happens from or against this target even when at max stacks).

internal/relic/cavern/championboxing/championboxing.go
```diff
@@ -0,0 +1,61 @@
...
+		Listeners: modifier.Listeners{
+			OnAfterAttack:        addAtkBuff,
+			OnAfterBeingAttacked: addAtkBuff,
+		},
+	})
+	modifier.Register(atkBuff, modifier.Config{
+		StatusType:        model.StatusType_STATUS_BUFF,
+		MaxCount:          5,
+		CountAddWhenStack: 1,
+	})
```
unleashurgeek, line 52: needs `Replace` or `ReplaceBySource` stacking behavior in order for modifier to stack

unleashurgeek (approved):

PR #214, opened 2023-09-08
Implement Bailu

RyanOld:


internal/character/bailu/eidolon.go
```diff
@@ -0,0 +1,60 @@
...
+			return
+		}
+		if e.Healer == c.id && e.Key == Skill {
+			c.engine.AddModifier(e.Target, info.Modifier{
+				Name:     E4,
+				Source:   c.id,
+				Stats:    info.PropMap{prop.AllDamagePercent: 0.1},
+				Duration: 2,
+			})
+		}
```
aaron-hwang, line 58: I believe instead of a `HealEnd` event subscription, you can instead use a listener for `OnAfterDealHeal` similar to the DM.  This is because the set of actions we care about in this case is limited to strictly actions dealt by Bailu.

internal/character/bailu/technique.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/simimpact/srsim/pkg/key"
+)
+
+// After Technique is used, at the start of the next battle,
+// all allies are granted Invigoration for 2 turn(s).
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	for _, char := range c.engine.Characters() {
+		c.addInvigoration(char, 2)
+	}
```
aaron-hwang, line 14: Duration of her technique's invigoration is affected by A4, so this needs to check for that

internal/character/bailu/trace.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		if e.Healer == c.id && e.OverflowHealAmount > 0 {
+			c.engine.AddModifier(e.Target, info.Modifier{
+				Name:     A2,
+				Source:   c.id,
+				Duration: 2,
+				Stats:    info.PropMap{prop.HPPercent: 0.1},
+			})
+		}
+	})
```
aaron-hwang, line 38: Similar to my comment about E4, I believe a listener is more suitable here since the set of actions we care about is a subset of all of Bailu's actions

internal/character/bailu/talent.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// remove invigoration once the heals are used up.
+	if state.healsLeft <= 0 {
+		mod.RemoveSelf()
+	}
+}
+
+// adds invigoration re-heal with independent heal counters to chars.
+// attached to char struct for easier reuse in ult and technique.
+func (c *char) addInvigoration(target key.TargetID, duration int) {
```
aaron-hwang, line 97: Instead of passing in the duration of invigoration as a paramter, I believe it would be easier to just have the check for A4 inside this function, since it affects invigorations from all parts of her kit

internal/character/bailu/talent.go
```diff
@@ -0,0 +1,120 @@
...
+// When an ally receives a killing blow, they will not be knocked down.
+// Bailu immediately heals the ally for 18% of Bailu's Max HP plus 480 HP.
+// This effect can be triggered 1 time per battle
+
+func init() {
+	modifier.Register(invigoration, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAfterBeingHitAll: healOnBeingHit,
+		},
+		Stacking:          modifier.Prolong,
```
aaron-hwang, line 34: In the dm, invigoration looks like it uses `ReplaceByCaster` instead of prolong

internal/character/bailu/eidolon.go
RyanOld: about using event subscription instead of a modifier with a listener, I do that because :
1. Tbh I tried using a listener and it ended up being longer (more lines of codes). I thought as long as I do the proper check (i.e. checking for caster is bailu, etc) it should be equivalent to a listener version.
2. Iirc my Seele PR also had some of these and it's still approved so I thought it's fine.
3. It increases the amount of modifiers. It's hardly an issue tho, just adds more stuff in the logs.
If you still want me to refactor it then I'll do it though dw.
I'll merge the early return conditions into 1 if statement while I'm at it too.

internal/character/bailu/eidolon.go
```diff
@@ -0,0 +1,59 @@
...
+		})
+	})
+
+	// E4 : Every healing provided by the Skill makes the recipient deal
+	// 10% more DMG for 2 turn(s). This effect can stack up to 3 time(s).
+	c.engine.Events().HealEnd.Subscribe(func(e event.HealEnd) {
+		// only add heal buff at E4+, healer is bailu, and it came from her skill casts.
+		if c.info.Eidolon < 4 || e.Healer != c.id || e.Key != Skill {
+			return
+		}
```
aaron-hwang, line 51: NIT: Can change to use `OnAfterDealHeal` listener instead, this is still correct but may be less performant? Probably not a huge deal....

internal/character/bailu/talent.go
```diff
@@ -0,0 +1,130 @@
...
+// After an ally with Invigoration is hit, restores the ally's HP
+// for 5.4% of Bailu's Max HP plus 144. This effect can trigger 2 time(s).
+// When an ally receives a killing blow, they will not be knocked down.
+// Bailu immediately heals the ally for 18% of Bailu's Max HP plus 480 HP.
+// This effect can be triggered 1 time per battle
+
+func init() {
+	modifier.Register(invigoration, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAfterBeingHitAll: healOnBeingHit,
```
aaron-hwang, line 32: Invigoration is `OnAfterBeingAttacked`, not `OnAfterBeingHitAll`. This current implementation has invigoration charges being drained and healing after each hit in an attack, and also includes DoT effects, which does not reflect in game behavior.

PR #227, opened 2023-09-17
Implement Dan Heng Imbibitor Lunae

interfacew:


internal/character/danhengimbibitorlunae/attack.go
```diff
@@ -0,0 +1,157 @@
...
+			HitRatio:     hitRatio,
+		})
+		c.engine.Attack(info.Attack{
+			Key:          Attack2,
+			HitIndex:     i,
+			Source:       c.id,
+			Targets:      c.engine.AdjacentTo(target),
+			DamageType:   model.DamageType_IMAGINARY,
+			AttackType:   model.AttackType_NORMAL,
+			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: enhancedAtk2[c.info.AttackLevelIndex()*3/19]},
```
Vessv, line 117: 3/19 Fraction is inside the array index.

internal/character/danhengimbibitorlunae/attack.go
```diff
@@ -0,0 +1,157 @@
...
+			HitRatio:     hitRatio,
+		})
+		c.engine.Attack(info.Attack{
+			Key:          Attack3,
+			HitIndex:     i,
+			Source:       c.id,
+			Targets:      c.engine.AdjacentTo(target),
+			DamageType:   model.DamageType_IMAGINARY,
+			AttackType:   model.AttackType_NORMAL,
+			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: enhancedAtk3[c.info.AttackLevelIndex()*9/25]},
```
Vessv, line 148: 9/25 Fraction is inside the array index

internal/character/danhengimbibitorlunae/ult.go
```diff
@@ -0,0 +1,69 @@
...
+			HitRatio:     hitRatio,
+		})
+		c.engine.Attack(info.Attack{
+			Key:          Ult,
+			HitIndex:     i,
+			Source:       c.id,
+			Targets:      c.engine.AdjacentTo(target),
+			DamageType:   model.DamageType_IMAGINARY,
+			AttackType:   model.AttackType_ULT,
+			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: ult[c.info.UltLevelIndex()*7/15]},
```
Vessv, line 47: 7/15 fraction is inside the Array Index.
```suggestion
			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: ult[c.info.UltLevelIndex()]*7/15},
```

internal/character/danhengimbibitorlunae/ult.go
```diff
@@ -0,0 +1,69 @@
...
+		Name:   Point,
+		Source: c.id,
+	})
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   Point,
+		Source: c.id,
+	})
+	// if E2,advanced forward 100% and 1 more point after ult
+	if c.info.Eidolon >= 2 {
+		c.engine.InsertAction(c.id)
```
Vessv, line 63: ImbibitorLunae E2 is 100% advance forward, not InsertAction. You need to use ModifyGaugeNormalized for this case, next is a suggestion but its not complete, i would advice you to look at other implementations to see how they did it and adapt it for this case.
```suggestion
		c.engine.ModifyGaugeNormalized(info.ModifyAttribute{
					Key:    ,
					Target: c.id,
					Source: c.id,
					Amount: -1,
				})
```

Vessv (changes requested):
For now this is what i could find in a bit of time, there might be others things.

internal/character/danhengimbibitorlunae/trace.go
```diff
@@ -0,0 +1,61 @@
...
+			Source: c.id,
+		})
+	}
+}
+func A6OnBeforeHit(mod *modifier.Instance, e event.HitStart) {
+	if e.Hit.Defender.IsWeakTo(model.DamageType_IMAGINARY) {
+		mod.SetProperty(prop.CritDMG, 0.24)
+	}
+}
+func A6OnAfterHit(mod *modifier.Instance, e event.HitEnd) {
```
Vessv, line 59: A6OnAfterHit func is not needed. A6OnBeforeHit only works for the hit so it will not affect Dan stats.

internal/character/danhengimbibitorlunae/skill.go
```diff
@@ -0,0 +1,82 @@
...
+		Source: c.id,
+	})
+	c.engine.InsertAction(c.id)
+}
+
+// add skill effect stack when attack
+func (c *char) AddSkill() {
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   SkillEffect,
+		Source: c.id,
```
Vessv, line 69: Outroar buff has `"LifeStepImmediately": true` in config. In this case, when you call AddModifier you need to add the `TickImmediately: true`
```suggestion
		Source: c.id,
		TickImmediately: true
```

internal/character/danhengimbibitorlunae/skill.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *char) initSkill() {
+	modifier.Register(EnhanceLevel, modifier.Config{
+		Stacking:          modifier.ReplaceBySource,
+		MaxCount:          3,
+		StatusType:        model.StatusType_UNKNOWN_STATUS,
+		CountAddWhenStack: 1,
+	})
+	// if E4, the effect has 1 more turn and can be refreshed
+	if c.info.Eidolon < 4 {
```
Vessv, line 30: Remove the conditional and just do one Register, when AddModifier is called for SkillEffect, check if E4 then Duration = 2 and finally add  Duration inside the info.Modifier. In this case the AddModifier for this modifier is in the AddSkill function.

This is how the Register would look like here.
```
modifier.Register(SkillEffect, modifier.Config{
			StatusType: model.StatusType_STATUS_BUFF,
			Stacking: modifier.ReplaceBySource,
			TickMoment: modifier.ModifierPhase2End
			MaxCount: 4,
			Duration: 1,
			Listeners: modifier.Listeners{
				OnAdd: SkillOnAdd,
			},
			CountAddWhenStack: 1,
		})
```

internal/character/danhengimbibitorlunae/skill.go
```diff
@@ -0,0 +1,82 @@
...
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   EnhanceLevel,
+		Source: c.id,
+	})
+	c.engine.InsertAction(c.id)
+}
+
+// add skill effect stack when attack
+func (c *char) AddSkill() {
+	c.engine.AddModifier(c.id, info.Modifier{
```
Vessv, line 67: Here is where the Outroar duration would be checked and if dan is E4 then duration = 2.
```suggestion
        outroarDuration := 1
	if c.info.Eidolon >= 4 {
	     outroarDuration = 2
	}
	
	c.engine.AddModifier(c.id, info.Modifier{
```
 then inside the info.Modifier you would set the duration to outroarDuration
``` go
Duration: outroarDuration
```

internal/character/danhengimbibitorlunae/skill.go
```diff
@@ -0,0 +1,82 @@
...
+			CritDMGBoost: skill[c.info.SkillLevelIndex()],
+		},
+	})
+}
+
+func SkillOnAdd(mod *modifier.Instance) {
+	state := mod.State().(skillState)
+	mod.SetProperty(prop.CritDMG, mod.Count()*(state.CritDMGBoost))
+}
+func SkillOnPhase2(mod *modifier.Instance) {
```
Vessv, line 80: func SkillOnPhase2 can be removed.

internal/character/danhengimbibitorlunae/attack.go
```diff
@@ -0,0 +1,178 @@
...
+			Targets:      []key.TargetID{target},
+			DamageType:   model.DamageType_IMAGINARY,
+			AttackType:   model.AttackType_NORMAL,
+			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: atk[c.info.AttackLevelIndex()]},
+			StanceDamage: 30,
+			EnergyGain:   20,
+			HitRatio:     hitRatio,
+		})
+		if i >= 3 {
+			c.AddSkill()
```
Vessv, line 92: OutRoar only works for Divine Spear or Fulgurant Leap (Lv 2 and 3 of basic), so this is conditional is better to remove it from here (Normal Basic)
```suggestion	
```

internal/character/danhengimbibitorlunae/attack.go
```diff
@@ -0,0 +1,178 @@
...
+			Targets:      []key.TargetID{target},
+			DamageType:   model.DamageType_IMAGINARY,
+			AttackType:   model.AttackType_NORMAL,
+			BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: enhancedAtk1[c.info.AttackLevelIndex()]},
+			StanceDamage: 60,
+			EnergyGain:   30,
+			HitRatio:     hitRatio,
+		})
+		if i >= 3 {
+			c.AddSkill()
```
Vessv, line 112: OutRoar only works for Divine Spear or Fulgurant Leap (Lv 2 and 3 of basic), so this is conditional is better to remove it from here (Lv 1 Basic)
```suggestion	
```

Vessv (commented):
Some other things i could find. As beofre, there are some suggestions, always check if the code makes sense lol(surely i didn't mess smt up).

internal/character/danhengimbibitorlunae/ult.go
```diff
@@ -0,0 +1,72 @@
...
+	Point            = "danhengimbibitorlunae-point"
+	Ult   key.Attack = "danhengimbibitorlunae-ult"
+	E2    key.Reason = "danhengimbibitorlunae-e2"
+)
+
+var ultHits = []float64{0.3, 0.3, 0.4}
+
+func init() {
+	modifier.Register(Point, modifier.Config{
+		StatusType:        model.StatusType_STATUS_BUFF,
```
Vessv, line 20: Squama Sacrosancta in datamine is  "StatusType": "Other", i think in this case the `StatusType:` need to be removed.
```suggestion
```

Vessv (commented):
This are some other things i found.
Another thing is, i think you need to check all the modifiers you implemented and only add `StatusType: ` when its a explicit buff or debuff. For example, i think the A4 modifier doesn't have that in datamine, so its prob better to remove it.

interfacew:
in the game, when dan unleash an enhanced attack, there is only 1 action.
but in this code, it used multi action(1 attack and 1-3 skill).

also in game, use the skill of dan won't trigger skill listener like yukong trace 103 to generate energy.

how can i fix this?

(sorry for my bad English)

internal/character/danhengimbibitorlunae/skill.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+type skillState struct {
+	CritDMGBoost float64
+}
+
+// enhance attack,has 3 type,use 1/2/3 skill point
+
+func (c *char) initSkill() {
+	modifier.Register(EnhanceLevel, modifier.Config{
```
Vessv, line 24: Keeping track of the `EnhanceLevel` via a modifier is not really needed. 

> so give char some like skillCount variable that you just increment every time skill is used and then set to 0 in Attack 
-Kyle

internal/character/danhengimbibitorlunae/ult.go
```diff
@@ -0,0 +1,74 @@
...
+	Point                  = "danhengimbibitorlunae-point" // MAvatar_DanHengIL_00_BPCostChange
+	UltPrimary  key.Attack = "danhengimbibitorlunae-ult-primary"
+	UltAdjacent key.Attack = "danhengimbibitorlunae-ult-adjacent"
+	E2          key.Reason = "danhengimbibitorlunae-e2"
+)
+
+var ultHits = []float64{0.3, 0.3, 0.4}
+
+func init() {
+	modifier.Register(Point, modifier.Config{
```
Vessv, line 20: Same as the `EnhancheLevel` Modifier. Keeping track of the `Point` via a modifier is not really needed. My suggestion instead would be to just use a variable , give there char some like `Point`variable that you just increment after ult then change it in Attack. Bscly what you are doing with with the `Point` modifier, but change to a variable and not a modifier.

Vessv:
> in the game, when dan unleash an enhanced attack, there is only 1 action. but in this code, it used multi action(1 attack and 1-3 skill).
> 
> also in game, use the skill of dan won't trigger skill listener like yukong trace 103 to generate energy.
> 
> how can i fix this?
> 
> (sorry for my bad English)

Hi, no worries, i am also a non native english speaker.
I also am not sure how 💀 
But from what Kyle told me

> from a sim perspective, we dont care about the internal distinction of multi actions like how it is implemented in game (at least not right now, though there is an argument to be made that we should). QQ is another case off the top of my head that has multiple skills technically. As long as the behavior is there, it doesn't materially matter what the action is since we do not have any "action keys"
The current implementation is definitely "incorrect" because of the whole "skill would count as an action" stuff, but that's okay because materially the implementation of DHIL should not change when that is fixed. It is okay to have the "incorrect" implementation merged it and then someone can follow up with the core changes necessary to support it
-Kyle

So don't worry about that for now, i think.

internal/character/danhengimbibitorlunae/eidolon.go
```diff
@@ -0,0 +1,30 @@
...
+
+// E1 increase talent max stack by 4 and gain 1 extra stack when hit
+// E2 100% forward and 1 more ult stack
+// E4 skill buff 1 more turn
+// E6 20% imaginary pen for attack3 when ally use ult, max 3 stack
+
+func init() {
+	// imaginary pen for attack3,change by ally ult count
+	modifier.Register(E6Effect, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
```
Vessv, line 21: E6 effect is a modifier that listen to OnBeforeHitAll and check if the attacks is lvl 3 then adds the res pen to that attack
```suggestion
		StatusType: model.StatusType_STATUS_BUFF,
		MaxCount: 3,
		CountAddWhenStack: 1,
		Stacking: modifier.ReplaceBySource,
		Listeners: modifier.Listeners{
			OnBeforeHitAll: E6PenOnBeforeHitAll,
			OnAfterAttack: E6PenOnAfterAttack,
		}
```
The beforehit function would look something like this.
```
func E6PenOnBeforeHitAll(mod *modifier.Instance, e event.HitStart) {
	if //check here if attack level is 3 {
		amt := mod.Count()*mod.State().(float64)
		e.Hit.Attacker.AddProperty(name, prop.ImaginaryPEN, amt)
	}
}
```
The afterattack function would look something like this.
```
func E6PenOnAfterAttack(mod *modifier.Instance, e event.AttackEnd) {
	if //check here if attack level is 3 {
		//remove E6Effect mod
	}
}
```

internal/character/danhengimbibitorlunae/eidolon.go
```diff
@@ -0,0 +1,30 @@
...
+	})
+}
+
+// count ally ult
+func (c *char) E6ActionEndListener(e event.ActionEnd) {
+	if e.AttackType == model.AttackType_ULT && e.Owner != c.id && c.E6Count < 3 {
+		c.E6Count++
```
Vessv, line 28: Here is where you need to add de E6Effect modifier, i think you dont need to E6Count bc the mod will only stack upto 3.
You will also need to check here if
```
if c.info.Eidolon >= 6 {
			//Add E6Effect modifier, the state would be 0.2 
		}
```

Vessv (commented):
This are the last changes i could find. 
This is just to replicate how the game does it and so it doesn't buff other attacks like Tingyun additional(pursued) attack.

Vessv (approved):
👍

PR #232, opened 2023-10-10
Implement hook

aaron-hwang:


internal/character/hook/technique.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const (
+	Technique = "hook-technique"
+)
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	techniqueTarget := c.engine.Retarget(info.Retarget{
+		Max:          1,
+		IncludeLimbo: false,
```
RyanOld, line 17: nit : you can safely remove this line as the default behavior is to exclude `onLimbo` targets.

RyanOld (commented):
Overall I think you should look into simplifying some of the codes by spotting repeated logics and where they're used and replace it with a general function that does the processing inside but that's just my opinion.
As long as it "just works" it should be fine.
It just makes things hard to review and debug later on if we find some issues.

internal/character/hook/skill.go
```diff
@@ -0,0 +1,125 @@
...
+func (c *char) Skill(target key.TargetID, state info.ActionState) {
+	if c.engine.HasModifier(c.id, SkillEnhancement) {
+		c.EnhancedSkill(target, state)
+	} else {
+		c.NormalSkill(target, state)
+	}
+
+	c.applySkillBurn([]key.TargetID{target})
+
+	c.engine.EndAttack()
```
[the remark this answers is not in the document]
RyanOld: Note : if you decide to add the `EndAttack`s below, the one here can safely be removed.

internal/character/hook/skill.go
```diff
@@ -0,0 +1,125 @@
...
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: skillNormal[c.info.SkillLevelIndex()],
+		},
+		EnergyGain:   30,
+		StanceDamage: 60,
+	})
+
+	if c.engine.HasBehaviorFlag(target, model.BehaviorFlag_STAT_DOT_BURN) {
+		c.talentProc(target)
+	}
```
RyanOld, line 41: ok, so about `EndAttack`, because the amount of hook's skill hits depends on various things like : 
- normal/enhanced skill
- talent is procced or not

we need to put `EndAttack` after the last hit inside every one of the above logics.
That is if we want to follow closely to the DM. I don't actually know how much of an effect this is, but the logs should at least get tidier. 
The first one is in an else here :
```suggestion
	} else {
	    c.engine.EndAttack()
	}
```
On normal skill and no talent proc, hook only hits once and the last hit is the above `Attack`.
Note : we don't want to just add `EndAttack` after normal/enhanced attack because there are a few logics after the attack ends.

internal/character/hook/skill.go
```diff
@@ -0,0 +1,125 @@
...
+			Target: c.id,
+			Source: c.id,
+			Amount: 5,
+		})
+		for _, t := range talentCanidates {
+			c.talentPursuedDamage(t)
+		}
+		c.applySkillBurn(c.engine.AdjacentTo(target))
+		c.talentHeal()
+	}
```
RyanOld, line 96: next is here in an else. If there are no talent candidates, last hit is the attacks on adjacent targets
```suggestion
	} else {
	    c.engine.EndAttack()
	}
```
...and that's it i think. i see you've already added the one in `talentPursuedDamage`.

RyanOld (approved):
I'm still not sure about the `init()` function you deleted. 
The only thing I know from watching other's codes is if you added a new modifier key, you gotta register it first inside an `init()` before using it.
But if you're sure about it then I'll trust you :)

PR #233, opened 2023-10-14
implement Yanqing

interfacew:


internal/character/yanqing/talent.go
```diff
@@ -0,0 +1,105 @@
...
+				}
+				for i, hitRatio := range followHits {
+					c.engine.Attack(info.Attack{
+						Key:          TalentAttack,
+						HitIndex:     i,
+						Source:       c.id,
+						Targets:      []key.TargetID{target},
+						DamageType:   model.DamageType_ICE,
+						AttackType:   model.AttackType_INSERT,
+						BaseDamage:   info.DamageMap{model.DamageFormula_BY_ATK: talentFollowRate[c.info.TalentLevelIndex()]},
```
aaron-hwang, line 76: nit: `talentFollowRate` might be a confusing name, could be interpreted as being similar to talent followup chance/rate

internal/character/yanqing/technique.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (c *char) Technique(target key.TargetID, state info.ActionState) {
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   TechEffect,
+		Source: c.id,
+	})
+}
+
+func TechOnHit(mod *modifier.Instance, e event.HitStart) {
+	if mod.Engine().HPRatio(e.Defender) > 0.5 {
```
aaron-hwang, line 35: should be `mod.Engine().HPRatio(e.Defender) >= 0.5`, since the effect condition is greater than or equal to.

internal/character/yanqing/trace.go
```diff
@@ -0,0 +1,46 @@
...
+
+func init() {
+	modifier.Register(A6, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+		Stacking:   modifier.Refresh,
+		Duration:   2,
+	})
+}
+
+func (c *char) A2Listener(e event.AttackEnd) {
```
aaron-hwang, line 25: This should be implemented as a listener for a modifier in my opinion, not as a subscriber to the `AttackEnd` event. Reasoning:
Current implementation listens for `AttackEnd`  events from ALL entities in the sim, not just yanqing, as subscribers listen for events from all sources that are of a specific event type. 
To fix this, you can instead use a modifier with the appropriate listener attached, as they inherently only care about events produced by their owner. You could also add a check as to whether or not the source of the attack is Yanqing, but I believe that implementation would be much messier. Former also matches more closely with the dm.

internal/character/yanqing/trace.go
```diff
@@ -0,0 +1,46 @@
...
+			Source:     e.Attacker,
+			Targets:    e.Targets,
+			DamageType: model.DamageType_ICE,
+			AttackType: e.AttackType,
+			BaseDamage: info.DamageMap{model.DamageFormula_BY_ATK: 0.3},
+		})
+	}
+}
+
+func (c *char) A6Listener(e event.HitEnd) {
```
aaron-hwang, line 38: same as above; We only care about hits Yanqing deals, so use a listener instead of an event subscription.

internal/character/yanqing/eidolon.go
```diff
@@ -0,0 +1,54 @@
...
+	if c.info.Eidolon >= 4 && e.NewHPRatio >= 0.8 {
+		c.engine.AddModifier(c.id, info.Modifier{
+			Name:   E4,
+			Stats:  info.PropMap{prop.IcePEN: 0.12},
+			Source: c.id,
+		})
+	} else {
+		c.engine.RemoveModifier(c.id, E4)
+	}
+}
```
aaron-hwang, line 47: See comments made in `trace.go` for A2 and A6 listeners

internal/character/yanqing/eidolon.go
```diff
@@ -0,0 +1,54 @@
...
+	E1 key.Attack = "yanqing-e1"
+)
+
+func init() {
+	modifier.Register(E4, modifier.Config{
+		StatusType: model.StatusType_STATUS_BUFF,
+	})
+}
+
+func (c *char) E1Listener(e event.AttackEnd) {
```
aaron-hwang, line 24: Same as E4/A2/A6

PR #235, opened 2023-10-16
Implement Serval

shinyfuwante:


internal/character/serval/talent.go
```diff
@@ -0,0 +1,49 @@
...
+
+const (
+	ServalTalentListener key.Modifier = "serval-talent-listener"
+	ServalTalent         key.Attack   = "serval-talent"
+)
+
+func init() {
+	modifier.Register(ServalTalentListener, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAfterHit: onAfterHit,
```
Vessv, line 20: Serval talent suscribes to OnAfterAttack not OnAfterHit

internal/character/serval/ult.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const Ult key.Attack = "serval-ult"
+
+func (c *char) Ult(target key.TargetID, state info.ActionState) {
+	if c.info.Eidolon >= 4 {
+		for _, trg := range c.engine.Enemies() {
+			c.engine.AddModifier(trg, info.Modifier{
```
Vessv, line 15: It only applies the modifier if the enemy doesnt have "Flag": "STAT_DOT_Electric"

internal/character/serval/ult.go
```diff
@@ -0,0 +1,43 @@
...
+		Targets:    c.engine.Enemies(),
+		DamageType: model.DamageType_THUNDER,
+		AttackType: model.AttackType_ULT,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: ult[c.info.UltLevelIndex()],
+		},
+		StanceDamage: 60.0,
+		EnergyGain:   5,
+	})
+	for _, trg := range c.engine.Enemies() {
```
Vessv, line 39: It only extends the DoT if the enemy has "Flag": "STAT_DOT_Electric"

internal/character/serval/trace.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// A6:
+//
+//	Upon defeating an enemy, ATK is increased by 20% for 2 turn(s).
+const (
+	A4 = "serval-a4"
+	A6 = "serval-a6"
+)
+
+func initTraces(c *char) {
```
Vessv, line 22: Better to at least make it a function of the char rather than its own function
```suggestion
func (c *char) initTraces() {
```
When you change it, remember to change the init in serval.go
c.initTraces()

Vessv (commented):
Some small things that need to be changed.

Vessv (approved):
Looks good

PR #245, opened 2023-12-22
Reactions

aaron-hwang:
WIP, need to add:
- Interface for modifiers, should contain function that will run custom event logic. This function will correspond to TriggerModifierCustomEvent in the DM. (Main purpose is for kafka implementation atp)
- Relevant states need to implement the function for aforementioned interface
- Notes: Name ideas for interface: Triggerable? CustomEventTriggerable?


internal/global/common/bleed.go
```diff
@@ -99,3 +99,13 @@ func breakBleedPhase1(mod *modifier.Instance) {
 		UseSnapshot:  true,
 	})
 }
+
+// Custom event trigger for bleed dots
+func (B BleedState) TriggerEvent(mod *modifier.Instance) {
```
unleashurgeek, line 104: nit: might not be clear to users what `TriggerEvent` means when reading character implementations (though not against the name)

Thinking about implementations, such as Sampo E4 or Kafka, these would also need to take in a `Ratio` which gets multiplied against the original damage percentage of the mod (IE: Sampo E4 is 4%). In the game implementation, this would be that `DOT_TriggerRatio`: https://damnae.github.io/stargazer/#/modifier-MCommon_Element_Electric/from/avatar-1005/@59d64be43a1da285cf22ba9be5ed90ef2b23f857

internal/global/common/custom_event_triggerable.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+type CustomEventTriggerable interface {
+	TriggerEvent()
+}
```
unleashurgeek, line 5: this should be specific to trigger dots/element modifiers and not some generic "any custom event" trigger

unleashurgeek (commented):
To test this implementation, can implement Sampo E4 as part of this PR: https://github.com/simimpact/srsim/issues/48

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
unleashurgeek, line 4: why was this changed?

internal/global/common/TriggerableDot.go
unleashurgeek, line 1: file name should be all lowercase with underscore separation (`triggerable_dot.go`)

internal/global/common/burn.go
```diff
@@ -83,9 +92,41 @@ func breakBurnPhase1(mod *modifier.Instance) {
...
+func (b BreakBurnState) TriggerDot(mod *modifier.Instance, ratio float64) {
+	// perform burn damage
+	mod.Engine().Attack(info.Attack{
+		Key:        BreakBurn,
+		Source:     mod.Source(),
+		Targets:    []key.TargetID{mod.Owner()},
+		AttackType: model.AttackType_DOT,
+		DamageType: model.DamageType_FIRE,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_BREAK_DAMAGE: b.BreakBaseMulti,
```
unleashurgeek, line 128: missing ratio?

internal/global/common/bleed.go
```diff
@@ -99,3 +99,38 @@ func breakBleedPhase1(mod *modifier.Instance) {
...
+func (b BleedState) TriggerDot(mod *modifier.Instance, ratio float64) {
+	// perform bleed damage
+	mod.Engine().Attack(info.Attack{
+		Key:        Bleed,
+		Source:     mod.Source(),
+		Targets:    []key.TargetID{mod.Owner()},
+		AttackType: model.AttackType_DOT,
+		DamageType: model.DamageType_PHYSICAL,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: b.DamagePercentage,
```
unleashurgeek, line 113: missing ratio?

internal/global/common/wind.go
```diff
@@ -78,7 +87,36 @@ func breakWindShearPhase1(mod *modifier.Instance) {
...
+
+func (w WindShearState) TriggerDot(mod *modifier.Instance, ratio float64) {
+	mod.Engine().Attack(info.Attack{
+		Key:        WindShear,
+		Source:     mod.Source(),
+		Targets:    []key.TargetID{mod.Owner()},
+		AttackType: model.AttackType_DOT,
+		DamageType: model.DamageType_WIND,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_ATK: w.DamagePercentage * mod.Count(),
```
unleashurgeek, line 105: missing ratio

internal/global/common/wind.go
```diff
@@ -78,7 +87,36 @@ func breakWindShearPhase1(mod *modifier.Instance) {
...
+
+func (w BreakWindShearState) TriggerDot(mod *modifier.Instance, ratio float64) {
+	mod.Engine().Attack(info.Attack{
+		Key:        BreakWindShear,
+		Source:     mod.Source(),
+		Targets:    []key.TargetID{mod.Owner()},
+		AttackType: model.AttackType_DOT,
+		DamageType: model.DamageType_WIND,
+		BaseDamage: info.DamageMap{
+			model.DamageFormula_BY_BREAK_DAMAGE: w.BreakBaseMulti * mod.Count(),
```
unleashurgeek, line 119: missing ratio

unleashurgeek (commented):
implementation lgtm. Would like to see Sampo E4 implemented and tested just to verify if it works (not 100% on if these function headers would be best as current state or against pointers): https://github.com/simimpact/srsim/issues/48.

internal/character/sampo/skill.go
```diff
@@ -25,9 +26,17 @@ func (c *char) Skill(target key.TargetID, state info.ActionState) {
...
+		// Wind dot (kit dots)
+		for _, dot := range c.engine.GetModifiers(target, common.WindShear) {
+			dot.State.(common.WindShearState).TriggerDot(dot, 0.06, c.engine, target)
+		}
+
+		// Break wind dots
+		for _, dot := range c.engine.GetModifiers(target, common.BreakWindShear) {
+			dot.State.(common.BreakWindShearState).TriggerDot(dot, 0.06, c.engine, target)
+		}
+	}
```
unleashurgeek, line 39: NIT: My FUD with this is you are checking by behavior flag but getting modifiers by name. Runs us into an odd situation for what if there is something that is `STAT_DOT_POISON` but not `WindShear` or `BreakWindShear`. Best if we had some way to `GetModifiersByBehaviorFlag` instead?

unleashurgeek (approved):

PR #252, opened 2023-12-23
Fixed stance break/reset event emit

agentyoda:


pkg/engine/attribute/modify.go
```diff
@@ -98,6 +99,19 @@ func (s *Service) SetStance(data info.ModifyAttribute) error {
 	}
 	attr := t.attributes
 
+	if data.Amount == 0 {
+		s.event.StanceBreak.Emit(event.StanceBreak{
+			Key:    data.Key,
+			Target: data.Target,
+			Source: data.Source,
+		})
+	} else if attr.Stance == 0 {
```
unleashurgeek, line 108: in `emitStanceChange` there is a check near the top of the function if the `prevS == newS` to exit early. Need to do that check prior to event emission to make sure we are not triggering break/reset on cases where no stance damage is happening (previous is 0 and new is 0)

agentyoda: ok, check added for no changes to stance

unleashurgeek (approved):

PR #253, opened 2023-12-24
Fix Stance Break/Reset

agentyoda:


pkg/engine/attribute/modify.go
```diff
@@ -98,32 +98,31 @@ func (s *Service) SetStance(data info.ModifyAttribute) error {
 		return fmt.Errorf("unknown target: %v", data.Target)
 	}
 	attr := t.attributes
+	prev := attr.Stance
+	attr.Stance = data.Amount
```
unleashurgeek, line 102: we need this assignment to happen after the event emission to fix the original issue of toughness modifier. Should have bounds logic against `data.Amount` and then after emit calls create prev and set the new stance value

agentyoda: ok, switched the assignments to the end

unleashurgeek (approved):

unleashurgeek (approved):

PR #255, opened 2023-12-26
fix program crash when a character skill on a dead ally

imring:
issue: https://discord.com/channels/845087716541595668/1101278107312336906/1189186773335691354

### DIFF
--- a/pkg/engine/target/character/execute.go
+++ b/pkg/engine/target/character/execute.go
@@ -40,11 +40,7 @@ func (mgr *Manager) ExecuteAction(id key.TargetID, isInsert bool) (target.Execut
 		return target.ExecutableAction{}, err
 	}
 
-	// checking target alive in advance is needed to avoid a program crash
-	// IsValid is required to exclude key.TargetEvaluator (First/LowestHP/etc)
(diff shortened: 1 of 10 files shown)


pkg/engine/target/character/execute.go
```diff
@@ -34,9 +34,25 @@ func (mgr *Manager) ExecuteAction(id key.TargetID, isInsert bool) (target.Execut
...
+	// checking target alive in advance is needed to avoid a program crash
+	// IsValid is required to exclude key.TargetEvaluator (First/LowestHP/etc)
+	isTargetDead := mgr.engine.IsValid(key.TargetID(act.TargetEvaluator)) && !mgr.attr.IsAlive(key.TargetID(act.TargetEvaluator))
+
+	if useSkill && !canUseSkill || isTargetDead {
+		useSkill = false
+		act, err = mgr.eval.DefaultAction(id)
+		if err != nil {
+			return target.ExecutableAction{}, err
+		}
```
unleashurgeek, line 52: should also check if this default action is a valid and alive target. If not, should just default to some arbitrary `TargetEvaluator` (`LowestHP` or something?)

imring: reworked: if the action has a dead target, there will be an error. added `is_alive` function to handle such cases.

unleashurgeek (approved):

PR #256, opened 2024-01-16
Implement Kafka

aaron-hwang:


internal/character/kafka/talent.go
```diff
@@ -0,0 +1,98 @@
...
+	c.canUseTalent = true
+}
+
+func (c *char) initTalent() {
+	c.engine.AddModifier(c.id, info.Modifier{
+		Name:   kafkaTalent,
+		Source: c.id,
+	})
+	c.engine.Events().AttackEnd.Subscribe(c.talentTrigger)
+	c.engine.Events().ActionEnd.Subscribe(c.talentAttack)
```
agentyoda, line 37: There's a bunch of linter errors that suggest these funcs (and funcs/vars they call) aren't being used. I'm not sure why, as this seems fine at first glance. Something to investigate, I suppose.

agentyoda (changes requested):
Thanks for your hard work on this! Left some comments, mostly relating to the lint failure.

internal/character/kafka/talent.go
agentyoda: On further review, it seems that these aren't being used b/c of the previous inits not being called. So if the other comments are fixed, these should be too.

agentyoda (approved):

agentyoda (approved):

agentyoda (approved):

PR #261, opened 2024-03-16
add black swan lightcone

alixdotsh:
i hope i did this correctly and committed correctly i never commit from cli 😭


alixdotsh:
nvm i realized i completely skipped a step let me cook

alixdotsh:
okay this is ready for review now

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,102 @@
...
+		Name:   remembrance,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectHitRate: ehrAmt},
+		State:  &modState,
+	})
+}
+
+func addProphetStack(mod *modifier.Instance, e event.HitEnd) {
+	state := mod.State().(*state)
+	if mod.Engine().HasBehaviorFlag(e.Defender, model.BehaviorFlag_STAT_DOT_ELECTRIC, model.BehaviorFlag_STAT_DOT_BURN, model.BehaviorFlag_STAT_DOT_BLEED) {
```
sky-larke, line 80: missing `model.BehaviorFlag_STAT_DOT_POISON` (wind shear)

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,102 @@
...
+func init() {
+	lightcone.Register(key.ReforgedRemembrance, lightcone.Config{
+		CreatePassive: Create,
+		Rarity:        5,
+		Path:          model.Path_NIHILITY,
+		Promotions:    promotions,
+	})
+	modifier.Register(remembrance, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnAfterHit: addProphetStack,
```
sky-larke, line 38: should be `OnBeforeHitAll` instead of `OnAfterHit`, prophet stacks are on the attacker, `OnBeforeHitAll` is also called  by the attacker, while `OnAfterHit` is called by the defender

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	engine.AddModifier(owner, info.Modifier{
+		Name:   remembrance,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectHitRate: ehrAmt},
+		State:  &modState,
+	})
+}
+
+func addProphetStack(mod *modifier.Instance, e event.HitEnd) {
```
sky-larke, line 78: `e event.HitEnd` should be `e event.HitStart` for `OnBeforeHitAll`

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,102 @@
...
+	modifier.Register(defShred, modifier.Config{
+		BehaviorFlags: []model.BehaviorFlag{
+			model.BehaviorFlag_STAT_DEF_DOWN,
+		},
+		Stacking:          modifier.ReplaceBySource,
+		StatusType:        model.StatusType_STATUS_BUFF,
+		MaxCount:          4,
+		CountAddWhenStack: 1,
+		Listeners: modifier.Listeners{
+			OnAdd: recalcDefShred,
```
sky-larke, line 59: change this to `OnAfterHit`, such as in here: https://github.com/simimpact/srsim/blob/59d22f4a344da43f147fb273d95c66e2496599e6/internal/lightcone/hunt/sleeplikethedead/sleeplikethedead.go#L32-L35
this way, you can check if the attack type is DoT before adding the DEF shred

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,102 @@
...
+		})
+	}
+}
+
+func recalcAtkBuff(mod *modifier.Instance) {
+	atkBuff := mod.State().(float64) * mod.Count()
+	mod.AddProperty(prop.ATKPercent, atkBuff)
+}
+
+func recalcDefShred(mod *modifier.Instance) {
```
sky-larke, line 99: following the above, I would add `e event.HitEnd` into `recalcDefShred`, and add a condition to check if the dmg is DoT before you add the shred

This example from Sleep Like the Dead is a good reference, except it has `AddModifier` where `AddProperty` would go in this case.
https://github.com/simimpact/srsim/blob/59d22f4a344da43f147fb273d95c66e2496599e6/internal/lightcone/hunt/sleeplikethedead/sleeplikethedead.go#L62-L63

sky-larke (changes requested):
Remember this: `in a single battle, only 1 stack of Prophet can be granted for each type of DoT`
Right now, there isn't a check for that, so it will just increase no matter what. I would separate the checks for each type of DoT. (it is also a good test to write)

alixdotsh:
I am going to bed .

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	engine.AddModifier(owner, info.Modifier{
+		Name:   remembrance,
+		Source: owner,
+		Stats:  info.PropMap{prop.EffectHitRate: ehrAmt},
+		State:  &modState,
+	})
+}
+
+func GetDotDamage(defender key.TargetID, flag model.BehaviorFlag) float64 {
```
sky-larke, line 85: doesn't this just return 0 every time? so sum is never updated

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,123 @@
...
+func addProphetStack(mod *modifier.Instance, e event.HitStart) {
+	state := mod.State().(*state)
+	sum := func(engine engine.Engine, _, sum float64) float64 {
+		for _, flag := range dotFlags {
+			if engine.HasBehaviorFlag(e.Defender, flag) {
+				sum += GetDotDamage(e.Defender, flag)
+			}
+		}
+		return sum
+	}
```
sky-larke, line 98: nit, just move back to a separate helper function instead of defining it here if you're doing it more than once

internal/lightcone/nihility/reforgedremembrance/reforgedremembrance.go
```diff
@@ -0,0 +1,123 @@
...
+
+	dotSum := sum(mod.Engine(), mod.Count(), 0)
+	if dotSum > 0 {
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:   atkBuff,
+			Source: mod.Owner(),
+			State:  state.atkBuff,
+		})
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
```
sky-larke, line 107: you still need to check if the damage type of the hit is DoT before you apply def shred to the hit

PR #195, opened 2023-08-12
Implement The Unreachable Side (LC)

RyanOld:

PR #267, opened 2024-07-29
Implement 10 Relics

kdovtdc:
* initial relic setup
* implement effects

complete:
* Cavern: Band, Eagle, Firesmith, Guard, Thief, Wastelander, Disciple, Messenger (with comment)
* Planar: Celestial (without Multi-wave support), Keel

incomplete:
* ~~Cavern: Disciple, Messenger~~
* ~~Planar: Keel~~

Notes:
* Disciple: ~~in onHPChange, it is assumed `e.Target` means source and compares to mod owner (= wearer; similar to #195) need to check whether this is correct~~ needs `e.Source` (from #281)
~~* Keel: how to handle `applied` and pointer logic~~ (idea from conversation on #124; once finished, can be used for Fleet impl since they are functionally almost identical)
* Messenger: missing logic to check whether at least one of the Ult's targets is an ally (intersection != nil) during `OnBeforeAction`, which requires the action's target list (`event.ActionStart` doesn't have it) -> will be left as a bug (issue created: https://github.com/simimpact/srsim/issues/283 )

### DIFF
--- a/internal/relic/cavern/wastelander/wastelander.go
+++ b/internal/relic/cavern/wastelander/wastelander.go
@@ -25,11 +25,13 @@ func init() {
 	relic.Register(key.WastelanderOfBanditryDesert, relic.Config{
 		Effects: []relic.SetEffect{
 			{
-				MinCount: 2,
-				Stats:    info.PropMap{prop.ImaginaryDamagePercent: 0.1},
+				MinCount:     2,
+				Stats:        info.PropMap{prop.ImaginaryDamagePercent: 0.1},
+				CreateEffect: nil,
 			},
 			{
 				MinCount: 4,
+				Stats:    nil,
 				CreateEffect: func(engine engine.Engine, owner key.TargetID) {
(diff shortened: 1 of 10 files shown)


internal/relic/cavern/wastelander/wastelander.go
```diff
@@ -0,0 +1,57 @@
...
+	})
+	modifier.Register(wastelander, modifier.Config{
+		Listeners: modifier.Listeners{
+			OnBeforeHitAll: onBeforeHitAll,
+		},
+	})
+}
+
+func onBeforeHitAll(mod *modifier.Instance, e event.HitStart) {
+	debuffCount := float64(e.Hit.Defender.StatusCount(model.StatusType_STATUS_DEBUFF))
```
aaron-hwang, line 50: float64 cast is unnecessary

aaron-hwang (commented):
Some minor formatting/general changes

PR #270, opened 2024-08-17
Implement Guinaifen

kdovtdc:
* skeleton + data
* implement stats, attack, skill, eidolons
* implement talent, traces, ult, technique

### DIFF
--- a/internal/character/guinaifen/ult.go
+++ b/internal/character/guinaifen/ult.go
@@ -25,7 +25,7 @@ func (c *char) Ult(target key.TargetID, state info.ActionState) {
 
 	// triggers all Burns on all enemies
 	for _, trg := range c.engine.Enemies() {
-		for _, dot := range c.engine.GetModifersByBehaviorFlag(target, model.BehaviorFlag_STAT_DOT_BURN) {
+		for _, dot := range c.engine.GetModifersByBehaviorFlag(trg, model.BehaviorFlag_STAT_DOT_BURN) {
 			dot.State.(common.TriggerableDot).TriggerDot(dot, ultDetonateBurn[c.info.UltLevelIndex()], c.engine, trg)
 		}
 	}


internal/character/guinaifen/skill.go
```diff
@@ -0,0 +1,96 @@
...
+			c.engine.AddModifier(target, info.Modifier{
+				Name:   common.Burn,
+				Source: c.id,
+				State: &common.BurnState{
+					DamagePercentage: skillBurn[c.info.SkillLevelIndex()],
+				},
+				Chance:   1,
+				Duration: 2,
+			})
+		}
```
aaron-hwang, line 64: if you want to keep it shorter you could instead do something like this (pseudocode):
```
burnMulti := skillBurn[c.info.SkillLevelIndex()]
if eidolon 2 && hasBurn, burnMulti += 0.4
c.engine.addModifier(etc)
```

internal/character/guinaifen/talent.go
```diff
@@ -0,0 +1,124 @@
...
+			mod.Engine().AddModifier(target, info.Modifier{
+				Name:   common.Burn,
+				Source: mod.Owner(),
+				State: &common.BurnState{
+					DamagePercentage: skillBurn[gui.SkillLevelIndex()],
+				},
+				Chance:   0.8,
+				Duration: 2,
+			})
+		}
```
aaron-hwang, line 122: same as comment on skill, could technically extract the actual burn application into a separate function and have both of them call it (with a parameter for the base chance to use)

internal/character/guinaifen/talent.go
```diff
@@ -0,0 +1,124 @@
...
+			Chance:            1,
+			Duration:          3,
+			MaxCount:          float64(maxStacks),
+			CountAddWhenStack: 1,
+		})
+	}
+}
+
+// calculate the received damage increase
+func FirekissonStack(mod *modifier.Instance) {
```
aaron-hwang, line 92: would capitalize the o for better readability

internal/character/guinaifen/ult.go
```diff
@@ -0,0 +1,37 @@
...
+	// triggers all Burns
+	burns := []key.Modifier{common.Burn, common.BreakBurn}
+	for _, trg := range c.engine.Enemies() {
+		for _, triggerable := range burns {
+			for _, dot := range c.engine.GetModifiers(trg, triggerable) {
+				dot.State.(common.TriggerableDot).TriggerDot(dot, ultDetonateBurn[c.info.UltLevelIndex()], c.engine, trg)
+			}
+		}
+	}
+
```
aaron-hwang, line 35: Can be refactored using the new `GetModifiersByBehaviorFlag` method in engine

aaron-hwang (approved):
LGTM

Issue #282, opened 2024-09-20
Himeko E6 is missing attack effect


kdovtdc:
Himeko E6 is iterating through targets without doing an attack against them:

https://github.com/simimpact/srsim/blob/6bff0380fdcd697b964fe3999d76dfb552f62aaa/internal/character/himeko/ult.go#L44-L53

PR #286, opened 2024-09-22
Fix Himeko E6 missing attack effect

kdovtdc:
closes https://github.com/simimpact/srsim/issues/282
* add E6 attack effect and key
* remove redundant filter


aaron-hwang (approved):
LGTM

PR #325, opened 2024-10-29
fix jingliu skill ratio, refactor e1

soloxcx:
- e1 should have its own key, doesn't need separate method


soloxcx:
drafting cause I want to refactor her global event subscriptions into listeners

PR #343, opened 2024-12-10
Canuse attack

### LINKED ISSUE #269: Add `CanUse` field for `Attack`, similar to `CanUse` for `Skill`
Needed for cases where certain conditions do not allow `Attack` to be performed (e.g. Jingliu's Spectral Transmigration state).

aaron-hwang:
Fixes #269

### DIFF
--- a/pkg/engine/target/character/execute.go
+++ b/pkg/engine/target/character/execute.go
@@ -83,6 +83,10 @@ func (mgr *Manager) ExecuteAction(id key.TargetID, isInsert bool) (target.Execut
 		return target.ExecutableAction{}, err
 	}
 
+	canUseAttack, err := mgr.engine.CanUseAttack(id)
+	if err != nil || !canUseAttack {
+		return target.ExecutableAction{}, err
+	}
 	return target.ExecutableAction{
 		Execute: func() {
 			char.Attack(primaryTarget, actionState{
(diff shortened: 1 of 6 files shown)


pkg/engine/target/character/execute.go
```diff
@@ -83,6 +83,10 @@ func (mgr *Manager) ExecuteAction(id key.TargetID, isInsert bool) (target.Execut
 		return target.ExecutableAction{}, err
 	}
 
+	canUseAttack, err := mgr.engine.CanUseAttack(id)
+	if err != nil || !canUseAttack {
+		return target.ExecutableAction{}, err
+	}
```
kdovtdc, line 89: might want to return a custom error message somewhere in case both `canUseSkill` and `canUseAttack` fail (which should never happen ingame)

kdovtdc (commented):
The CanUse part looks fine to me – but I think with the way `ExecuteAction` was designed, we might also need to address the "fallback" logic now.

At the moment, if Skill is called, the `ExecuteAction` checks for `canUseSkill` first, and if it fails, falls back to using Attack (which is desired).
Now, if Attack is called but fails at `canUseAttack` (e.g. a Jingliu in her Enhanced state), ideally we want to fall back to Skill, otherwise we won't have anything to return.

For that, I wouldn't do a big refactor for now, as this part might (or might not) see an overhaul someday.

Issue #345, opened 2025-01-11
Modify Max SP count


aaron-hwang:
Currently, there is no way for character implementations to modify the maximum amount of SP (skill points) available in the simulation. It can be seen below that we currently check a hard value of a cap of 5 whenever we modify SP. There is a character known as Sparkle who modifies the SP cap, so it will need to be implemented so her implementation can be completed.

https://github.com/simimpact/srsim/blob/f2a58eaefcbdd7f89ce3c5c9d72dcf2e9089cdc8/pkg/engine/attribute/modify.go#L193-L202

PR #346, opened 2025-01-12
Sparkle

aaron-hwang:
Only missing functionality should be #345


internal/character/sparkle/skill.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/simimpact/srsim/pkg/engine/info"
+	"github.com/simimpact/srsim/pkg/engine/modifier"
+	"github.com/simimpact/srsim/pkg/engine/prop"
+	"github.com/simimpact/srsim/pkg/key"
+	"github.com/simimpact/srsim/pkg/model"
+)
+
+const (
+	SparkleSkillBuff = "SparkleSkillBuff"
+	Dreamdiver       = "dreamdiver"
```
kdovtdc: naming-wise, I would probably call the first-applied mod "dreamdiver" and the a4 extension "dreamdiver-a4-extend" or something

internal/character/sparkle/skill.go
```diff
@@ -0,0 +1,116 @@
...
+func init() {
+	modifier.Register(SparkleSkillBuff, modifier.Config{
+		Stacking:   modifier.ReplaceBySource,
+		CanDispel:  true,
+		StatusType: model.StatusType_STATUS_BUFF,
+		Listeners: modifier.Listeners{
+			OnAdd:    addActualBuff,
+			OnRemove: A4Extend,
+		},
+		TickMoment: modifier.ModifierPhase1End,
```
kdovtdc: this one should be default tick moment

internal/character/sparkle/skill.go
```diff
@@ -0,0 +1,116 @@
...
+		Duration:   1,
+	})
+
+	modifier.Register(Dreamdiver, modifier.Config{
+		Stacking:   modifier.Replace,
+		CanDispel:  true,
+		StatusType: model.StatusType_STATUS_BUFF,
```
kdovtdc, line 32: this one should have `ModifierPhase1End` tick moment

internal/character/sparkle/skill.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+	})
+}
+
+func A4Extend(mod *modifier.Instance) {
+	sparkleinfo, _ := mod.Engine().CharacterInfo(mod.Source())
+	if sparkleinfo.Traces["102"] {
+		mod.Engine().AddModifier(mod.Owner(), info.Modifier{
+			Name:   Dreamdiver,
+			Source: mod.Source(),
```
kdovtdc, line 113: `TargetSelf` as caster should mean source is mod.Owner() I believe

internal/character/sparkle/talent.go
```diff
@@ -0,0 +1,109 @@
...
+type talentState struct {
+	DmgPercentPerStack float64
+	isE2               bool
+}
+
+func (c *char) initTalent() {
+	c.engine.Events().SPChange.Subscribe(c.talent)
+	c.engine.Events().ModifierAdded.Subscribe(c.adjustTalentBuff)
+	c.engine.Events().ModifierRemoved.Subscribe(c.revertTalentBuff)
+	//TODO: Modify maximum sp count once the api for doing so is defined.
```
kdovtdc, line 37: mentioned in #347

internal/character/sparkle/talent.go
```diff
@@ -0,0 +1,109 @@
...
+					isE2:               c.info.Eidolon >= 2,
+				},
+			})
+		}
+	}
+}
+
+func adjustBuff(mod *modifier.Instance) {
+	curstate := mod.State().(talentState)
+	// TODO: Maybe refactor the two buffs to be tied to the same modifier like in dm? end res is same but might be more readable
```
kdovtdc, line 60: I'd agree with that
the Dmg Bonus mod also needs to be removed if the Buff is removed with OnRemove

internal/character/sparkle/talent.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/simimpact/srsim/pkg/engine/prop"
+)
+
+const (
+	SparkleTalent = "red-herring"
+)
+
+func init() {
+	modifier.Register(SparkleTalent, modifier.Config{
+		Stacking: modifier.ReplaceBySource,
```
kdovtdc, line 17: missing `StatusType: model.StatusType_STATUS_BUFF,` and `CanDispel: true,`

internal/character/sparkle/traces.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		for _, char := range c.engine.Characters() {
+			charinfo, _ := c.engine.CharacterInfo(char)
+			atkBuff := 0.15
+			if charinfo.Element == model.DamageType_QUANTUM {
+				atkBuff += buffMap[quacount]
+			}
+			c.engine.AddModifier(char, info.Modifier{
+				Name:   A6,
```
kdovtdc, line 57: is this modifier is missing registration?

internal/character/sparkle/ult.go
```diff
@@ -0,0 +1,90 @@
...
+
+const (
+	Cipher = "sparkle-cipher"
+	Ult    = "sparkle-ult"
+)
+
+func init() {
+	modifier.Register(Cipher, modifier.Config{
+		Stacking:   modifier.ReplaceBySource,
+		CanDispel:  false,
```
kdovtdc: nit: can omit this if false (default)

internal/character/sparkle/ult.go
```diff
@@ -0,0 +1,90 @@
...
+					Source:   c.id,
+					Duration: 1,
+					Stats: info.PropMap{
+						prop.CritDMG: sparkleCdmg * proportion,
+					},
+				})
+			}
+		}
+	}
+
```
kdovtdc, line 66: I think we need to correctly match what dm does, as it applies first buff or second buff depending on what it finds on the randomly selected ally

kdovtdc (commented):
review for the moment

internal/character/sparkle/traces.go
aaron-hwang: added

internal/character/sparkle/talent.go
aaron-hwang: thinking about this a little more, i think it would be a little easier / potentially more performant to separate the buffs into two modifiers.
Reasoning;
If i keep the two buffs in one modifier (like in the dm), it would require having a check for every hit in every attack of whether or not Sparkle is e2 (potentially costly? either way seems like an unncessary perf hit). If we seperate the two modifiers, we can guarantee that the e2 modifier is only added when sparkle is e2. Alternatively we could keep them both in one modifier, and then carry over some state in the modifier of whether or not sparkle was/is e2. Might be less readable though.

kdovtdc: should be okay as two modifiers (though it probably won't be that much better in performance since the E2 check happens with every `adjustBuff` call anyways + new memory required for another mod? I'm not an expert in this so idk what's better)

internal/character/sparkle/talent.go
```diff
@@ -0,0 +1,112 @@
...
+			Source:    mod.Source(),
+			CanDispel: false,
+			Count:     mod.Count(),
+		})
+	}
+}
+
+func (c *char) adjustTalentBuff(e event.ModifierAdded) {
+	if e.Modifier.Name == Cipher && e.Modifier.Source == c.id {
+		for _, char := range c.engine.Characters() {
```
kdovtdc, line 74: in this subscription, this should only happen for the Cipher mod owner (instead of looping over all chars); otherwise Sparkle applying Cipher to 4 chars would cause every char to get the mod applied 4 times

internal/character/sparkle/talent.go
```diff
@@ -0,0 +1,112 @@
...
+					isE2:               c.info.Eidolon >= 2,
+				},
+			})
+		}
+	}
+}
+
+func (c *char) revertTalentBuff(e event.ModifierRemoved) {
+	if e.Modifier.Name == Cipher && e.Modifier.Source == c.id {
+		for _, char := range c.engine.Characters() {
```
kdovtdc, line 92: similar to above, each char should have the buff reverted for themselves separately (should be how it also works ingame)

internal/character/sparkle/ult.go
```diff
@@ -0,0 +1,99 @@
...
+			})
+		} else if c.engine.HasModifier(char, DreamdiverExtension) {
+			c.engine.AddModifier(char, info.Modifier{
+				Name:     DreamdiverExtension,
+				Source:   c.id,
+				Duration: 1,
+				Stats: info.PropMap{
+					prop.CritDMG: sparkleCdmg * proportion,
+				},
+			})
```
kdovtdc, line 96: this would only reapply the mod if the char already has said mod; what we need instead is take the char determined from Retarget method and apply that mod to all other allies (= excluding self)

PR #347, opened 2025-01-14
Add Max SP logic

### LINKED ISSUE #345: Modify Max SP count

kdovtdc:
* add Max SP to Manager
* add methods `MaxSP`: returns Max SP
* add `ModifyMaxSP`: modifies Max SP similarly to `ModifySP` with `Amount` field; no handling of error cases (closes #345)
* add `MaxSPChange` event + emit
* update `ModifySP` logic to use `maxsp` instead of fixed number
* rerun scripts


aaron-hwang (approved):
lgtm

